Issue #9, opened 2019-12-08
Option for using cert-manager for TLS / clarify docs


davidkarlsen:
https://github.com/anchore/anchore-charts/blob/master/stable/anchore-admission-controller/templates/init-ca/init-ca-script.yaml seems to have some manual jobs to be applied for generating some TLS certificates, and it will require internet access as it's based on some downloads, apt etc. This is a bit of a hassle /tricky in air-gapped environments where you might need to go through a proxy etc.

Depending on what the requirements to this certificate (what is it used for / why is it - could maybe be more smooth by using cert-manager for certificate issuing and rotation? Maybe it's even better to handle this outside of the chart and simply refer to an existing secret (which could be generated by any means - including cert-manager).

Background in slack: https://anchorecommunity.slack.com/archives/C4PJFNEEM/p1575834841399400

zhill:
I'm all for simplifying this. I tried at one point using the CA abilities of Sprig but wasn't quite able to get it working on the apiserver side where the CA seemed to be not configured quite correctly. I think the requirements are: autogen one if not provided, and enable user to easily add their own if desired. Beyond that, I'm all for best-practices here.

FOTempel:
Soo, what's the progress on this? Requiring internet access for a single use init container stopped me dead in my tracks with deploying this chart. I understand that the certificates are needed to make the API server communicate with the admissioncontroller. Understandable. But it would be great if I could just inject my own certificates as a secret or something without the need for that very large init container and complex scripting that goes around it. I'd wager most organizations security conscious enough to even consider an admission controller already have their own ways and methods to deal with certificates. A simple switch in the chart to disable the init container as a quick "fix" would already go a long way, I can manage my own required k8s resources if needed (some documentation on that would also be appeciated but I can reverse engineer it if needed).

davidkarlsen:
I have a few PRs in (and one merged already) to clean things up. Also CI is added so it is faster and safer to move around. Follow the PRs which are active.

zhill:
Yes, this is an in-progress work. Understand the need and will ensure it gets added. Thanks for a bunch of great work by @davidkarlsen on this!

zhill:
Looking into it further, I'm not sure we even need the APIService component at all that can greatly simplify the chart. I think as long as a valid caBundle is present in the service and in the webhook client config, that the webhook can be called cleanly. I'm investigating this and any feedback appreciated.

FOTempel:
[...]

PR #13, opened 2019-12-13
add webhook - fixes #10

### LINKED ISSUE #10: automate webhook registration and improve policy choice
https://github.com/anchore/anchore-charts/blob/master/stable/anchore-admission-controller/files/get_validating_webhook_config.sh seems to contain some manual steps for registering the webhook. This should be documented in the README.md

Additionally could this simply be templated and handled by helm alongside the rest of the chart?

Further I think it makes sense to set the [failurePolicy](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy) to `Ignore` in order to avoid the control-plane malfunctioning if the controller should start failing.

davidkarlsen:
Signed-off-by: David J. M. Karlsen <david@davidkarlsen.com>

---
This change is [Reviewable](https://reviewable.io/reviews/anchore/anchore-charts/13)


stable/anchore-admission-controller/templates/webhook.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - operations:
+    - CREATE
+    apiGroups:
+    - ""
+    apiVersions:
+    - "*"
+    resources:
+    - pods
+  failurePolicy: {{ .Values.apiService.webhook.failurePolicy }}
+  namespaceSelector:
```
zhill: For this, how about adding a section in the values.yaml that lets a user insert a selector section verbatim. That would allow support for other types of selectors as well. Maybe a key 'selectors' that has its content added into the webhook object.

stable/anchore-admission-controller/values.yaml
```diff
@@ -15,6 +15,11 @@ service:
   externalPort: 443
 
 apiService:
+  webhook:
+    enabled: true
```
zhill: The expectation here is that to remove the configuration a user manually deletes it, or updates the values and redeploys? I can't recall if helm will remove the config if enabled = false on upgrade.

davidkarlsen: behaviour is configurable - depends on what you want.
If you want the webhook to be retained, then https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource can be used.

davidkarlsen:
Do you have any CI?
It could be automated with github-actions / kind / helm-tooling-actions like this: https://github.com/evryfs/helm-charts/commit/0aa28bdb22f3471a3c09265082765f97b38dae3e/checks?check_suite_id=358671120

zhill:
> Do you have any CI?
> It could be automated with github-actions / kind / helm-tooling-actions like this: https://github.com/evryfs/helm-charts/commit/0aa28bdb22f3471a3c09265082765f97b38dae3e/checks?check_suite_id=358671120

Yeah, that's a good call. We also use CircleCI so that's an option. I'll look into it.

zhill:
I'll be getting to this soon. The holidays in the US have delayed my availability but should be getting it fully reviewed and able to merge soon. Thanks for your patience @davidkarlsen

davidkarlsen:
I wonder if the ca-bundle is really needed https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#webhook-configuration, "URL" _Here is an example of a mutating webhook configured to call a URL (and expects the TLS certificate to be verified using system trust roots, so does not specify a caBundle):_

davidkarlsen:
I’ll Get a fresh squashed PR which passes dco once the other one is merged...

davidkarlsen:
OK - clean patch applied.
Can you test this in a cluster? I am not sure that the caBundle actually needs to be set for successful use&installation.

davidkarlsen:
@zhill did you have a chance to look at this? Does it work OOTB w/o any caBundle?

davidkarlsen:
@zhill ?

zhill:
> @zhill ?

I've done some very initial testing that it does install OOMB w/o caBundle, but need further testing to ensure it works as expected. I'll try to get to that later tonight so we can get this resolved soon. Thanks @davidkarlsen !

zhill:
I'm going to go ahead and merge this. I'm seeing some issues with the apiservice selection but I think that is not part of this change since it is setup in the init-ca script run during the install init-ca job.
```
kc get apiservice/v1beta1.admission.anchore.io
NAME                           SERVICE                                        AVAILABLE                  AGE
v1beta1.admission.anchore.io   default/admctrl-anchore-admission-controller   False (MissingEndpoints)   13s

```

davidkarlsen:
Ah - init-ca - we set out to fix that in #9. But is that really needed?

davidkarlsen:
OK, found a good description here: https://github.com/elithrar/admission-control#generating-tls-certificates

davidkarlsen:
And also https://www.scottguymer.co.uk/post/kubernetes-mutating-webhook-configuration/, lets fix this in #9

PR #61, opened 2020-08-31
Update chart to enterprise v2.4.0

Btodhunter:


Btodhunter:
The failed tests were run before the new images were published, so the install couldn't be completed.

PR #62, opened 2020-09-01
Helm chart Updated to pass secrets via environment variables.

Kishorb:
Hi, I updated to pass the DB credentials as env variables rather than k8s secrets. Our intension is to directly inject secrets into k8s pods from the Hashicorp vault using mutating webhooks ( Ref: https://banzaicloud.com/blog/inject-secrets-into-pods-vault-revisited).
We already deployed vault in the same k8s cluster and create secrets in the vault. In order to mutating webhook to inject secrets into pods, all pods should contain certain annotations, and we have to pass the env vault in vault format

for example, with this version code if you pass variables like below, then it will set below annotations on all pods and eventually when the pod starts vault will inject secrets into them as env variables (Ref: https://github.com/banzaicloud/bank-vaults/tree/master/charts/vault-secrets-webhook)

anchoreGlobal:
  annotations:
    vault.security.banzaicloud.io/vault-addr: "https://vault:8200"
    vault.security.banzaicloud.io/vault-role: "secrets_access" 
    vault.security.banzaicloud.io/vault-tls-secret: "vault-tls"
  extraEnv:
    # https://docs.anchore.com/current/docs/engine/engine_installation/configuration/network_proxies/
    - name: http_proxy
      value: http://corporateproxy:3128
    - name: https_proxy
      value: http://corporateproxy:3128
    - name: no_proxy
      value: "localhost,127.0.0.1"
    - name: ANCHORE_DB_HOST
      value: <db endpoint>
    - name: ANCHORE_DB_PASSWORD
      value: "vault:secret/data/anchore/db_password#ANCHORE_DB_PASSWORD"
    - name: ANCHORE_CLI_PASS
      value: "vault:secret/data/anchore/db_password#ANCHORE_DB_PASSWORD"
    - name: ANCHORE_APPDB_URI
      value: "vault:secret/data/anchore/appdb_uri#ANCHORE_APPDB_URI"
    - name: ANCHORE_ADMIN_PASSWORD
      value: "vault:secret/data/anchore/admin_password#ANCHORE_ADMIN_PASSWORD"

Regards,
Kishor.

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: anchore-engine
-version: 1.9.0
+version: 1.9.3
 appVersion: 0.8.1
 description: Anchore container analysis and policy evaluation engine service
 keywords:
--- a/stable/anchore-engine/templates/analyzer_deployment.yaml
+++ b/stable/anchore-engine/templates/analyzer_deployment.yaml
@@ -35,8 +35,11 @@ spec:
         {{- with .Values.anchoreGlobal.labels }}
         {{ toYaml . | nindent 8 }}
         {{- end }}
-      {{- with .Values.anchoreAnalyzer.annotations }}
       annotations:
+      {{- with .Values.anchoreGlobal.annotations }}
+        {{ toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.anchoreAnalyzer.annotations }}
         {{ toYaml . | nindent 8 }}
       {{- end }}
     spec:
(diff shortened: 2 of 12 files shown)


Btodhunter:
@Kishorb Thanks for the update! This looks like good functionality to add to our chart. Can you please sign your commits so that the DCO check passes. You'll also need to bump the chart version again since another PR has been merged since you submitted this one. I'm going to do some testing of these changes and will approve/merge after all those things have been completed.

Kishorb:
@Btodhunter - could you please check now. Thank you.

Btodhunter:
@Kishorb thanks for pushing those updates! I just noticed one more thing. It looks like templates/enterprise_feeds_upgrade_job.yaml doesn't have its env vars overwritten by .Values.inject_secrets_via_env

Btodhunter (changes requested):
@Kishorb thanks for pushing those updates! I just noticed one more thing. It looks like templates/enterprise_feeds_upgrade_job.yaml doesn't have its env vars overwritten by .Values.inject_secrets_via_env

Kishorb:
@Btodhunter - please check now. I hope it is good enough for you to merge it now.

Btodhunter:
@Kishorb LGTM. Thanks!

PR #64, opened 2020-09-03
Allow configuration of probe timeouts, failure/success thresholds

dakaneye:


dakaneye:
Test failures appear to be environmental, looks like an image failed to pull? Or perhaps the output isn't showing?

Btodhunter:
@dakaneye thanks for the update! Can you bump the chart version one more time? I just merged a PR. 

Also, if we're going to make the probes configurable, I'd like to see that applied to every deployment template not just the API.

Btodhunter (approved):

PR #75, opened 2020-10-17
Add extraEnv support for init pod

kroussou:
Allows passing environment variables to init-ca pod. Useful when k8s nodes are behind a proxy.

### DIFF
--- a/stable/anchore-admission-controller/values.yaml
+++ b/stable/anchore-admission-controller/values.yaml
@@ -80,7 +80,9 @@ credentials: {}
   # - username: user2
   #   password: password2
 
-# Define custom environment variables to pass to init-ca pod
-extraEnv: []
-  # - name: FOO
-  # value: "bar"
+# Settings related to init-ca pod
+initCA:
+  # Define custom environment variables to pass to init-ca pod
+  extraEnv: []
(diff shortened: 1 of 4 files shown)


stable/anchore-admission-controller/values.yaml
```diff
@@ -79,3 +79,8 @@ credentials: {}
   #   password: password1
   # - username: user2
   #   password: password2
+
+# Define custom environment variables to pass to init-ca pod
+extraEnv: []
```
zhill, line 84: I think to make this more clear that it is specifically for the init-ca pod, it should have something clear in the name. Perhaps either nest it under an 'init-ca' object or `initCaExtraEnv` as the name so that it isn't confusing to folks who might expect it to be in the other pod as well.

zhill:
Thanks for another PR @kroussou !

stable/anchore-admission-controller/values.yaml
kroussou: @zhill Makes sense. I'll do the update.

kroussou: @zhill Done 😀

zhill (approved):
lgtm

PR #85, opened 2020-11-16
Add passing of dbConfig.engineArgs in enterprise configmap

rabadin:
The global DB configuration is used in this file but the value of
'engineArgs' wasn't passed. This change adds 'engineArgs' to the
DB configuration.


Btodhunter (approved):
LGTM! Thanks for the contribution @rabadin

PR #93, opened 2020-12-14
Added the ability to point mulitple anchore deployments at a single enterprise feeds service

blang9238:


stable/anchore-engine/templates/engine_configmap.yaml
```diff
@@ -73,7 +73,9 @@ data:
           # Sync github data if available for GHSA matches
           github: {{ default "true" .Values.anchoreGlobal.syncGithub }}
       {{- if and .Values.anchoreEnterpriseGlobal.enabled .Values.anchoreEnterpriseFeeds.enabled }}
```
Btodhunter, line 75: Leaving the check for the feeds url inside the `if and` condition creates a situation where the url will not get populated if `.Values.anchoreEnterpriseFeeds.enabled=false`, which think will generally be the case when specifying an external URL for the feeds service.

This would allow the URL to be set when enterprise and/or enterprise feeds is disabled.
```
    {{- if and .Values.anchoreEnterpriseGlobal.enabled .Values.anchoreEnterpriseFeeds.enabled }}
      {{- if .Values.anchoreGlobal.internalServicesSsl.enabled }}
      url: "https://{{- template "anchore-engine.enterprise-feeds.fullname" . }}:{{- .Values.anchoreEnterpriseFeeds.service.port }}/v1/feeds"
      {{- else }}
      url: "http://{{- template "anchore-engine.enterprise-feeds.fullname" . }}:{{- .Values.anchoreEnterpriseFeeds.service.port }}/v1/feeds"
      {{- end }}
      ssl_verify: {{ .Values.anchoreGlobal.internalServicesSsl.verifyCerts }}
      client_url:
      token_url:
    {{- else if .Values.anchoreEnterpriseFeeds.url }}
      url: "{{- .Values.anchoreEnterpriseFeeds.url }}"
      ssl_verify: {{ .Values.anchoreGlobal.internalServicesSsl.verifyCerts }}
      client_url:
      token_url:
    {{- else }}
      client_url: "https://ancho.re/v1/account/users"
      token_url: "https://ancho.re/oauth/token"
      anonymous_user_username: anon@ancho.re
      anonymous_user_password: pbiU2RYZ2XrmYQ
    {{- end }}
```

Btodhunter (changes requested):

PR #94, opened 2020-12-14
Added the ability to point mulitple anchore deployments at a single enterprise feeds service

blang9238:


Btodhunter (approved):
Thanks @blang9238! LGTM

PR #97, opened 2020-12-16
Allow upgrade Jobs to have Annotations

adawalli:
In certain deployments, it is important to have control over the
annotations associated with an Upgrade Deployment. For example, if
primarily Fargate Pods are used in AWS, the upgrade jobs might need to
be run in EC2 context in order to speed up the helm upgrade process.


Btodhunter (approved):
Thanks for the update @adawalli! LGTM

Issue #100, opened 2021-01-07
Fixup Permissions in Scratch Directory when using AWS EFS Storage


adawalli:
When using AWS' EFS storage through the k8s CSI driver, the volume is mounted and owned by `root:root` in the deployed pods. 

k8s does allow enforcement of group ownership through their [fsGroup](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod) parameter.

Unfortunately, as documented by [here](https://github.com/kubernetes/examples/issues/260), [here](https://stackoverflow.com/questions/46769339/how-to-change-permission-of-mapped-volume-in-kubernetes-docker), and [here](https://docs.aws.amazon.com/efs/latest/ug/accessing-fs-nfs-permissions.html), NFS permissions do not get corrected properly by **fsGroup**

As documented in those threads, this problem can be resolved with some **initContainer** magic. I will submit a PR to address this.

Btodhunter:
fixed with pr #101

PR #101, opened 2021-01-07
Allow permission fixup to scratch space

adawalli:
NFS Mounted Scratchspaces do not get properly permissioned according to
the k8s Volume 'fsGroup' property. The most common solution for fixing
this is to use an initContainer to fix the permissions before the target
deployed POD runs. This solution implements just such an initContainer.
By default, this change has no impact on existing deployments and is "on
by request only" feature.

This PR addresses https://github.com/anchore/anchore-charts/issues/100

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.10.7
+version: 1.10.8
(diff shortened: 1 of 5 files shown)


adawalli:
before you merge this, let me test out EFS Access points to see if they are a viable alternative

Btodhunter (approved):
Thanks @adawalli! LGTM

PR #102, opened 2021-01-08
Use k8s secret for SAML shared secret

### LINKED ISSUE #60: SAML secret should be put in kubernetes secret
Currently in the `stable/anchore-engine/templates/engine_configmap.yaml`, the saml secret is placed via a hard-coded values.yaml value. This should be able to be placed in a kubernetes secret.

```
    # Locations for keys used for signing and encryption. Only one of 'secret' or 'public_key_path'/'private_key_path' needs to be set. If all are set then the keys take precedence over the secret value
    # Secret is for a shared secret and if set, all components in anchore should have the exact same value in their configs.
    keys:
     >>> secret: {{ .Values.anchoreGlobal.saml.secret }}<<<
      {{- with .Values.anchoreGlobal.saml.publicKeyName }}
      public_key_path: /home/anchore/certs/{{- . }}
      {{- end }}
      {{- with .Values.anchoreGlobal.saml.privateKeyName }}
      private_key_path: /home/anchore/certs/{{- . }}
      {{- end }}
```

Btodhunter:
Uses k8s secrets for storing SAML secret, which is mounted as ENV vars for running pods. Replacing hard coded value into the config.

fixes #60 

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.11.0
+version: 1.11.1
 appVersion: 0.9.0
 description: Anchore container analysis and policy evaluation engine service
 keywords:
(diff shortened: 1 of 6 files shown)


robertp (approved):
This looks to me like it should work, but I'm not well versed in SAML.

adawalli:
This looks good to me - no obvious errors. I am looking forward to this one!

Issue #52, opened 2020-08-07
Allow specifying database endpoint as an existing secret


Btodhunter:
Right now the ANCHORE_DB_HOST environment variable is specified in the configmap and cannot be overridden with the .Values.anchoreGlobal.existingSecret value

flickerfly:
Really, it'd be nice to have all the database connection information (username, port) in one secret and then specify just that in existingSecret. Also, I'd want to make sure that the .Values.anchoreGlobal.existingSecret implementation also cooresponded to the other uses of existingSecret keys throughout the chart. 

What would be even better is if I could simply mark something like `.Values.anchoreGlobal.existingSecrets: true` and all the charts would assume I'm using a pre-defined named secret. Just make them something like anchore-custom-ui, anchore-custom-feed, anchore-custom-engine, then document that and I'll create those ahead of time. Tracking down all this stuff to implement this securely is driving me to want to give up and take a less secure and easy way.

PR #106, opened 2021-01-12
Make secrets more consistent

### LINKED ISSUE #92: Inconsistent Secret Naming Convention in Helm Charts
In almost all cases within your helm charts, you will see secrets with the option to pull from the existingSecret secret, like this:

https://github.com/anchore/anchore-charts/blob/e0e753ca42d6b5ec041d9923e60062e5bf5a0c54/stable/anchore-engine/templates/api_deployment.yaml#L110

However, in the case of the Enterprise Feeds, you will see the following template used instead

https://github.com/anchore/anchore-charts/blob/e0e753ca42d6b5ec041d9923e60062e5bf5a0c54/stable/anchore-engine/templates/enterprise_feeds_deployment.yaml#L94

This is extra confusing since the .feedsDbPassword is an option that is documented here: https://github.com/anchore/anchore-charts/blob/e0e753ca42d6b5ec041d9923e60062e5bf5a0c54/stable/anchore-engine/templates/secrets.yaml

Can you please fix the chart so that feeds can pull from existingSecret?

Btodhunter:
Use environment variables in the anchore config.yaml file for feeds DB connection endpoint. This allows the feeds service to use the same mechanism for existing secrets as other deployments in the chart.

fixes #92 & #52

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.11.1
+version: 1.11.2
 appVersion: 0.9.0
 description: Anchore container analysis and policy evaluation engine service
 keywords:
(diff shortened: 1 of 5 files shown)


robertp (approved):
yep, looks good!

PR #107, opened 2021-01-12
use init job to set env vars in UI configuration. 

### LINKED ISSUE #91: Anchore Helm Chart for Enterprise UI Ignores "existingSecret" Postgres Password
Throughout nearly all of the Helm Chart for Anchore, the existingSecret is properly respected for pulling secrets. However, in the case of the Enterprise UI, this existing secret is completely ignored for the Postgres Password, and .Values.postgresql.postgresPassword is used (which won't be the right password).

https://github.com/anchore/anchore-charts/blob/master/stable/anchore-engine/templates/enterprise_ui_config_secret.yaml#L55 

Can the UI deployment please read this password from the existing secret instead of insecurely reading only from values?

Btodhunter:
This allows for using existing k8s secrets like all other deployments.

Not convinced an init job is the right solution for this. ENV var interpolation is on the development agenda for the UI, however there is not a timeline for this feature implementation yet.

fixes #91 

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.11.1
+version: 1.11.3
 appVersion: 0.9.0
 description: Anchore container analysis and policy evaluation engine service
 keywords:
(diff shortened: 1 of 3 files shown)


holt:
@Btodhunter I'll create a story for `3.1` that addresses this requirement in the app.

Issue #108, opened 2021-01-13
Bug in SAML existingSecret Logic


adawalli:
https://github.com/anchore/anchore-charts/pull/102 Introduced a really nice fix for pushing the SAML secrets into k8s secret store.  However, there is a tiny bug that requires `.Values.anchoreGlobal.saml.secret` to be set to any non-nil value even when `useExistingSecret==true`. You can set `.Values.anchoreGlobal.saml.secret` to anything you want as long as it doesn't evaluate to an empty/non-nil condition.

https://github.com/anchore/anchore-charts/blob/master/stable/anchore-engine/templates/engine_configmap.yaml#L102-L104

PR #109, opened 2021-01-13
Allow configmaps to read saml.useExistingSecret flag

adawalli:
There is currently a bug that requires the saml.secret value to be set
regardless of the value of saml.useExistingSecret. This changes
addresses this issue.

fixes: https://github.com/anchore/anchore-charts/issues/108

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.11.2
+version: 1.11.3
 appVersion: 0.9.0
 description: Anchore container analysis and policy evaluation engine service
 keywords:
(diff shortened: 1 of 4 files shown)


Btodhunter (approved):
Thanks @adawalli! Very good catch. LGTM.

Btodhunter (changes requested):
@adawalli sorry, I may have approved this a little too quickly. Just had it dawn on me that I'd rather have the SAML secret _always_ get setup as an ENV var in secrets.yaml

To accomplish that, I think the conditional should be:
```
{{- if or .Values.anchoreGlobal.saml.secret .Values.anchoreGlobal.saml.useExistingSecret }}
secret: ${ANCHORE_SAML_SECRET}
{{- end }}
```

adawalli:
ah, that makes sense - working on it now

adawalli:
There you go @Btodhunter !

Btodhunter (approved):
LGTM! Thanks @adawalli

PR #110, opened 2021-01-21
UI allow existing secrets

### LINKED ISSUE #91: Anchore Helm Chart for Enterprise UI Ignores "existingSecret" Postgres Password

Btodhunter:
This PR moves env var secrets into service specific k8 secrets (ui, feeds, engine). It also sets up the UI to use ENV vars for all endpoints that contain secrets. This will allow users to specify an existingSecret in their values file for setting secrets outside of their helm deployment.

fixes #91

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.11.3
+version: 1.11.4
 appVersion: 0.9.0
 description: Anchore container analysis and policy evaluation engine service
 keywords:
--- a/stable/anchore-engine/templates/_helpers.tpl
+++ b/stable/anchore-engine/templates/_helpers.tpl
@@ -139,3 +139,17 @@ Return Anchore Engine default admin password
     {{- randAlphaNum 32 -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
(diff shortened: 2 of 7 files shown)


robertp (dismissed):
LGTM

adawalli:
I pulled down your MR branch and tested locally - this PR is perfect! I was able to drop my `--set` override of the postgres password for feeds (after I realized I needed to add a secret for anchoreEnterpriseFeeds)

Good work @Btodhunter !!

adawalli:
One final note - the only real issue I am running into is that it does seem that the APP DB URI needs to be built by hand into the secret if going the existing secret route - correct?

robertp (approved):

Btodhunter:
@adawalli You are correct. That is a limitation of the UI. At this point it does not handle interpolation of ENV vars in the config, just full overrides. That functionality is in the works for 3.1 though.

PR #111, opened 2021-01-22
Use the API hostname so that the notifications service can be accessi…

dakaneye:
…ble from the catalog service


Btodhunter (approved):
Thanks @dakaneye! LGTM

PR #119, opened 2021-02-04
DON'T MERGE: Runtime Inventory Improvements' configurations

dakaneye:
Relates to both https://github.com/anchore/enterprise/issues/482 and https://github.com/anchore/enterprise/issues


stable/anchore-engine/templates/engine_configmap.yaml
```diff
@@ -256,4 +256,11 @@ data:
         ssl_cert: "/home/anchore/certs/{{- .Values.anchoreGlobal.internalServicesSsl.certSecretCertName }}"
         ssl_key: "/home/anchore/certs/{{- .Values.anchoreGlobal.internalServicesSsl.certSecretKeyName }}"
         {{- end }}
-
+    runtime_inventory:
+      kubernetes:
+        report_anchore_cluster:
```
Btodhunter, line 261: These config values can just use the defaults in the values.yaml file. In the config I used the default function for "hidden" config values. I want them toggle-able, but they should only be toggled when you know exactly what you're doing, so I don't put them in the values file to keep people from changing it willy nilly. This is not considered helm best practice, but there are things that you only want people to change when you explicitly tell them to.

Btodhunter: Alternatively, we could stop defaulting configs and put them all into the values file like helm best practices say. We'd just want to comment them heavily so people know the implications.

dakaneye: @Btodhunter Got it, I made the update, does it look right now?

Btodhunter: Yeah that looks great!

stable/anchore-engine/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.12.0
+version: 1.12.1
```
Btodhunter, line 3: This is gonna need to be bumped again, another PR snuck in before this one.

stable/anchore-engine/templates/engine_configmap.yaml
```diff
@@ -256,4 +257,11 @@ data:
         ssl_cert: "/home/anchore/certs/{{- .Values.anchoreGlobal.internalServicesSsl.certSecretCertName }}"
         ssl_key: "/home/anchore/certs/{{- .Values.anchoreGlobal.internalServicesSsl.certSecretKeyName }}"
         {{- end }}
-
```
Btodhunter, line 259: Is this a config that should only exist for enterprise?

dakaneye: @Btodhunter yeah

stable/anchore-engine/Chart.yaml
dakaneye: fixed!

robertp (approved):
LGTM, note that I have not tested these changes

PR #121, opened 2021-02-05
Found it update comments

found-it:
Modified the comments associated with the enterprise secrets so the name of the secret matches the name in the chart


Btodhunter (approved):
lgtm!

PR #123, opened 2021-02-10
If the feeds service is disabled, also disable the feeds upgrade job

Btodhunter:


robertp (approved):
LGTM

PR #125, opened 2021-03-05
Add checksums for configs

mjnagel:
Currently when a config change is made in the Helm values the running pods will not reflect that change due to the way configs are mounted in as volumes or env. Post-restart the pods reflect the changes but this process is not automated.

This PR aims to solve this issue by making use of [Helm checksums](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) to roll pods when a config value changes that would affect any mounted or sourced configmaps/secrets.

Changes:
- Splits secrets out into separate files - This is done to minimize rolls done based on checksum of the secret, rolls should only happen if the specific secret being used is modified
- Adds checksum annotations in deployments for any configmaps or secrets that are used as a volume or `envFrom`

This works for almost all config changes except the admin password changing. That requires a change to the value in the database as well. I think that would be best handled by something like the db upgrade jobs, it's pretty simple for cases with non-hashed passwords to accomplish with a k8s job that runs `psql` and updates the value but without understanding the hashing method/process better I don't know how to implement this *cleanly* for hashed passwords.

Signed-off-by: Micah Nagel <micah.nagel@parsons.com>


Btodhunter:
@mjnagel Thanks for the contribution! This is a great addition to the chart. I'll approve & merge this as soon as I'm able to do a little testing on it. 

Also, we had to bump the chart version for a engine/enterprise patch release, so can you bump the chart version one more time?

mjnagel:
@Btodhunter done! Let me know if you hit any issues, the only thing I know of right now is the admin password piece mentioned in the first comment, which isn't a big problem if the intent is to not allow changing password via a chart upgrade.

Btodhunter (approved):
LGTM! I don't think the chart needs to handle changing the anchore admin password, so this is great. Thanks @mjnagel.

PR #132, opened 2021-03-23
check for nil return value from helm lookup function

### LINKED ISSUE #131: Upgrades with --dry-run fail
When trying to perform a helm upgrade with the `--dry-run` flag, the following logic causes the upgrade to fail - https://github.com/anchore/anchore-charts/blob/master/stable/anchore-engine/templates/anchore_admin_secret.yaml#L5-L15

The helm lookup function requires connecting to the k8's API, which is not allowed during a dry run, so it returns a nil value. The nil return value causes the following index function to error out, causing the upgrade failure.

Btodhunter:
fixes #131 

This is not a perfect fix, if the user let helm create a random admin password by not setting `.Values.anchoreGlobal.defaultAdminPassword`, and then performs an upgrade with `--dry-run` the dry run output will show that the ANCHORE_ADMIN_PASSWORD value will change upon upgrade. This will not be the case when the actual upgrade is applied though, helm will lookup & reuse the password found in the existing k8s secret.

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.12.6
+version: 1.12.7
 appVersion: 0.9.3
 description: Anchore container analysis and policy evaluation engine service
 keywords:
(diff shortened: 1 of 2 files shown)


dakaneye (approved):
LGTM!

robertp (approved):
This worked for me

PR #133, opened 2021-03-31
WIP: add gitops method for db automation

bhearn7:
This PR aims to solve a potential use case for gitops teams who are using external postgres databases. Two new jobs were introduced to utilize postgres credentials (from values.yaml) from two new secrets in order to automate the following:
- selecting OR creating a database
- selecting a database user and altering that user to have database privileges OR creating a database user and altering that user to have database privileges

################

**Notes:**
- `pre-install,pre-upgrade` Helm hooks were used instead of `post-install,post-upgrade` so that the anchore jobs/components wouldn't error on "$DATABASE does not exist"
- Weights were used so the secrets (-5) would be created before the jobs (-4)
- Separate secrets and jobs were used for the main-db and feeds-db so users could use different database endpoints and credentials in values.yaml

**Concerns:**
- the jobs use a postgresql image that would need to be specified in values.yaml
- the jobs use imagePullSecrets that would need to be specified in values.yaml

**Known Issues:**
- if users set their database names with a hyphen (ie. anchore-feeds) the psql commands in the jobs will fail
- when upgrading in certain scenarios (ie. engine to enterprise, internal postgres to external dbs,) the engine-upgrade job will fail and restart 3-4 times until the anchore pods are ready, then the job will complete and the helm release will be ready

Signed-off-by: bhearn7 <blakeeh723@gmail.com>

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.12.7
+version: 1.12.8
 appVersion: 0.9.3
 description: Anchore container analysis and policy evaluation engine service
 keywords:
--- a/stable/anchore-engine/templates/anchore_db_ensure_job.yaml
+++ b/stable/anchore-engine/templates/anchore_db_ensure_job.yaml
@@ -1,4 +1,5 @@
-{{- if and (not .Values.postgresql.enabled) .Values.anchoreEnterpriseGlobal.enabled }}
+{{- if hasKey .Values.postgresql "enabled" }}
+{{- if (not .Values.postgresql.enabled) }}
 # Job to sync db and db user with external postgres for Anchore's primary data store
 apiVersion: batch/v1
 kind: Job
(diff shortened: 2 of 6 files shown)


Btodhunter (changes requested):
lets add another conditional to these jobs/secrets that defaults to false, because not everyone using an external DB will want the helm chart to try and create the db/users.
I'm thinking something like this would be sufficient:
```
{{- if .Values.postgresql.createExternalDB }}
```

Another thing to note - the value set in `.Values.postgresl.postgresPassword` will have to be the root/admin password of the external DB system to create those users/dbs. Same goes for `.Values.postgresql.postgresUser`. However, most people probably won't want that to also be the password/username of the application users. So we will want to either create random passwords for the application users & hardcode the engine/feeds db user names, or add values for those application usernames/passwords.

bhearn7:
@Btodhunter made some changes per your recommendations. I'm not sure why the lint check is failing on L686 for trailing space

templates:
- added condititonal "if createExternalDbAssets" for both secrets and both jobs
- added APPLICATION_USER and APPLICATION_PASSWORD in both secrets
- added conditional "if imagePullSecrets" for both jobs
- changed PGUSER to APPLICATION_USER in both jobs
- changed PGPASSWORD to APPLICATION_PASSWORD in both jobs

values:
- modified values.yaml to include `image`, `imagePullPolicy`, and `imagePullSecretName` for .postgresql and .anchore-feeds-db
- modified values.yaml to include `createExternalDbAssets`, `applicationUser`, and `applicationPassword` for .postgresql and .anchore-feeds-db

Btodhunter:
Closing this PR as outside the scope of our chart.

PR #136, opened 2021-04-15
Use env for tokens

Btodhunter:
GitHub & MSRC tokens should be stored in a secret and exposed as ENV vars to the pods.


robertp (approved):

robertp (dismissed):

robertp (approved):
LGTM

PR #147, opened 2021-05-25
Add OOB inventory details

dakaneye:
Add OOB runtime inventory configuration details to the helm chart.
This will allow anchore, running in the helm chart, to be able to collect inventory information from the cluster that it's in, out of the box (without user intervention required, by default)

### DIFF
--- a/stable/anchore-engine/values.yaml
+++ b/stable/anchore-engine/values.yaml
@@ -549,12 +549,21 @@ anchoreCatalog:
   tolerations: []
   affinity: {}
 
-  runtime_inventory:
+  runtimeInventory:
     # This setting tells Anchore how long an image can be missing from an inventory report before it is removed from
     # The working set. Note: The image will still have a historical record in the reports service, subject to data history
     # constraints as part of that service.
     # Note: if a runtime inventory image's digest is also in anchore for regular image analysis, it won't be removed.
-    image_ttl_days: 1
+    imageTTLDays: 1
+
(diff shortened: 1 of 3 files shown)


stable/anchore-engine/values.yaml
```diff
@@ -556,6 +556,16 @@ anchoreCatalog:
     # Note: if a runtime inventory image's digest is also in anchore for regular image analysis, it won't be removed.
     image_ttl_days: 1
 
+    # Since Anchore is running in Kubernetes, we can collect runtime inventory data out of the box
+    kubernetes:
```
zhill, line 560: Is this nesting necessary? Generally for chart configs you want to keep things as flat as possible. Also generally want to use camel casing in values.yaml unless keys are directly exposed to the app config.

Can it be something like:
```
deploymentInventory:
  enabled: true
  reportedClusterName: anchore-k8s
  namespaces:
    - all

zhill (changes requested):
slight changes in names and try to flatten a bit for simpler values.yaml, but nothing major

stable/anchore-engine/values.yaml
dakaneye: fixed!

robertp (approved):
I tested that a helm install w/these changes works, but not the runtime inventory features themselves.

zhill (approved):
lgtm. I have not manually tested that it works, just reviewed the updates.

PR #148, opened 2021-05-27
Add role and binding if OOB runtime inventory config is enabled

dakaneye:

### DIFF
--- a/stable/anchore-engine/templates/catalog_deployment.yaml
+++ b/stable/anchore-engine/templates/catalog_deployment.yaml
@@ -255,6 +255,25 @@ spec:
     app: {{ template "anchore-engine.fullname" . }}
     component: {{ $component }}
 
+{{- if .Values.anchoreCatalog.serviceAccount.enabled }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.anchoreCatalog.serviceAccount.name }}
+  labels:
+    {{- with .Values.anchoreCatalog.serviceAccount.labels }}
+    {{ toYaml . | nindent 4 }}
+    {{- end }}
+    {{- with .Values.anchoreGlobal.labels }}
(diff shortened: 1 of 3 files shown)


robertp (approved):
This LGTM but I also want @Btodhunter's eyes on it.

stable/anchore-engine/templates/catalog_deployment.yaml
```diff
@@ -254,3 +254,29 @@ spec:
...
+- apiGroups: [""]
+  resources: ["pods","namespaces"]
+  verbs: ["get", "watch", "list"]
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: catalog-cluster-read-only-binding
+subjects:
```
Btodhunter, line 274: @dakaneye I believe that this is going to need a service account to bind to, which the chart doesn't currently create.

Btodhunter (changes requested):

stable/anchore-engine/templates/catalog_deployment.yaml
Btodhunter: The chart does let you specify a service account to run deployments as, but nothing in the chart actually creates the service account. I think that if we have a thing that _requires_ a service account, we should create it in the chart.

dakaneye: I added one for the catalog specifically!

Btodhunter (changes requested):

stable/anchore-engine/values.yaml
```diff
@@ -535,6 +535,13 @@ anchoreCatalog:
     annotations: {}
     labels: {}
 
+    # These are the service account parameters for the catalog deployment
```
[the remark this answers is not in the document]
dakaneye: Fixed!

PR #149, opened 2021-06-09
Update admission controller chart to v0.3.0 with support for v0.3.0

zhill:
Updates the controller chart with some new configuration options like webhook rule customization.

Also bumps the chart and app versions to 0.3.0 (coincidence they align).


Btodhunter (approved):
lgtm!

PR #150, opened 2021-06-16
Updates comment about role

hev:


stable/kai/values.yaml
```diff
@@ -112,7 +112,7 @@ kai:
   anchore:
     url:
 
-    # Note: the user specified must be an admin user or have full-control, or read-write RBAC permissions
+    # Note: reccommend using the inventory-agent role
```
zhill, line 115: Typo :)
```suggestion
    # Note: recommend using the inventory-agent role
```

zhill (changes requested):
fix typo

PR #155, opened 2021-08-12
Add option to force cloud_sql_proxy connect to instance's private IP

rogeriobastos:
If an CloudSql instance has both public and private IP addresses, by default cloud_sql_proxy try to connect first to public IP address. We need to set `-ip_address_types=PRIVATE` to cloud_sql_proxy use the private address.

Reference: https://github.com/GoogleCloudPlatform/cloudsql-proxy#-ip_address_typespublicprivate


Btodhunter:
Closing as stale - a fix for this is implemented in #232

PR #160, opened 2021-08-18
Update engine 0101

### LINKED ISSUE #158: enabling/disabling feeds in values.yaml does not work unless booleans are quoted
enabling/disabling feeds in values.yaml does not work unless booleans are quoted

Broken Example:
```
anchoreEnterpriseFeeds:
  alpineDriverEnabled: false
  amazonDriverEnabled: false
  centosDriverEnabled: false
  debianDriverEnabled: false
  olDriverEnabled: false
  rhelDriverEnabled: false
  nvdDriverEnabled: false
  npmDriverEnabled: false
  gemDriverEnabled: false
  nvdv2DriverEnabled: false
  vulndbDriverEnabled: false
  msrcDriverEnabled: false
```

Working Example:
```
anchoreEnterpriseFeeds:
  alpineDriverEnabled: "false"
  amazonDriverEnabled: "false"
  centosDriverEnabled: "false"
  debianDriverEnabled: "false"
  olDriverEnabled: "false"
  rhelDriverEnabled: "false"
  nvdDriverEnabled: "false"
  npmDriverEnabled: "false"
  gemDriverEnabled: "false"
  nvdv2DriverEnabled: "false"
  vulndbDriverEnabled: "false"
  msrcDriverEnabled: "false"
```

### LINKED ISSUE #159: Cannot disable or enable MSRC Feed
Edit: Cannot enable MSRC Feed either.

Cannot disable MSRC Feed without causing exception to be raised:

values.yaml
```
anchoreGlobal:
  image: docker.io/anchore/anchore-engine-dev:v0.10.1-rc1
  defaultAdminPassword: foobar
  imagePullPolicy: Always
  syncVulnDB: false
  syncMSRC: false

anchoreEnterpriseGlobal:
  enabled: true
  image: docker.io/anchore/enterprise-dev:latest
  imagePullPolicy: Always


anchoreAnalyzer:
  enableOwnedPackageFiltering: true

anchoreEnterpriseFeeds:
  alpineDriverEnabled: "false"
  amazonDriverEnabled: "false"
  centosDriverEnabled: "false"
  debianDriverEnabled: "false"
  olDriverEnabled: "false"
  rhelDriverEnabled: "false"
  nvdDriverEnabled: "false"
  npmDriverEnabled: "false"
  gemDriverEnabled: "false"
  nvdv2DriverEnabled: "false"
  vulndbDriverEnabled: "false"
  msrcDriverEnabled: "false"
  ```

Logs:
```
[MainThread] [anchore_engine.configuration.localconfig/validate_config()] [ERROR] variable overrides found in configuration file that are unset (${ANCHORE_MSRC_KEY})
[MainThread] [anchore_enterprise_manager.cli.service/start()] [ERROR] Error: cannot load local configuration: invalid configuration: details - variable overrides found in configuration file that are unset (${ANCHORE_MSRC_KEY})
```

### PR DESCRIPTION
**Pull request by Btodhunter**
Update Anchore Engine, Enterprise & Enterprise-UI images to the latest version. 
Fix issue with booleans inside of `default` functions
Remove MSRC token configuration

fixes #158
fixes #159

### DIFF
--- a/stable/anchore-engine/templates/enterprise_feeds_configmap.yaml
+++ b/stable/anchore-engine/templates/enterprise_feeds_configmap.yaml
@@ -94,51 +94,51 @@ data:
           workspace_preload_file: "/workspace_preload/data.tar.gz"
         # If api_only is set to true, the service will not update feed data in the system.
         # API end points will be functional and serve feed data if any is available.
-        api_only: {{ default "false" (.Values.anchoreEnterpriseFeeds.apiOnly | toString) }}
+        api_only: {{ default "false" (.Values.anchoreEnterpriseFeeds.apiOnly | quote) }}
         drivers:
           # Configuration section for drivers collecting and processing feed data.
           # All drivers are enabled by default unless explicitly disabled.
           amzn:
-            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.amazonDriverEnabled | toString) }}
+            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.amazonDriverEnabled | quote) }}
           alpine:
-            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.alpineDriverEnabled | toString) }}
+            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.alpineDriverEnabled | quote) }}
           centos:
-            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.centosDriverEnabled | toString) }}
+            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.centosDriverEnabled | quote) }}
           debian:
-            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.debianDriverEnabled | toString) }}
+            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.debianDriverEnabled | quote) }}
           ol:
-            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.olDriverEnabled | toString) }}
+            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.olDriverEnabled | quote) }}
           ubuntu:
-            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.ubuntuDriverEnabled | toString) }}
+            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.ubuntuDriverEnabled | quote) }}
           rhel:
-            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.rhelDriverEnabled | toString) }}
+            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.rhelDriverEnabled | quote) }}
           nvddb:
-            enabled: {{ default "false" (.Values.anchoreEnterpriseFeeds.nvdDriverEnabled | toString) }}
+            enabled: {{ default "false" (.Values.anchoreEnterpriseFeeds.nvdDriverEnabled | quote) }}
           # npm and gem drivers are explicitly disabled out of the box
           npm:
-            enabled: {{ default "false" (.Values.anchoreEnterpriseFeeds.npmDriverEnabled | toString) }}
+            enabled: {{ default "false" (.Values.anchoreEnterpriseFeeds.npmDriverEnabled | quote) }}
           gem:
             # rubygem data comes packaged as a PostgreSQL dump file. gem driver loads the pg dump and normalizes the data.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: stable/anchore-engine/templates/enterprise_feeds_configmap.yaml [hunk 1]
```diff
@@ -94,54 +94,51 @@ data:
...
             {{- with .Values.anchoreEnterpriseFeeds.msrcWhitelist }}
             whitelist:
               - {{ . }}
             {{- end }}
-        {{- end }}
-        {{- if .Values.anchoreEnterpriseFeeds.githubDriverEnabled }}
           github:
-            enabled: true
+            enabled: {{ default "false" (.Values.anchoreEnterpriseFeeds.githubDriverEnabled | quote) }}
+          {{- if .Values.anchoreEnterpriseFeeds.githubDriverEnabled }}
```
Vijay-P, line 137: Should this not be `{{- if eq .Values.anchoreEnterpriseFeeds.githubDriverEnabled "true"}}`, since we are dealing in strings and not actual booleans?

Vijay-P: Tested with `helm template -g ./ -f ./engine-rc-values.yaml -s templates/enterprise_feeds_configmap.yaml`

Btodhunter: Good observation. You're right that we're comparing strings not booleans, however if the value in the values.yaml file isn't quoted, then you'll get an error - `error calling eq: incompatible types for comparison`. We could fix this by using:
```
{{- if eq (.Values.anchoreEnterpriseFeeds.githubDriverEnabled | quote) "true" }}
```

However that isn't the convention that is used extensively throughout the chart, so I'm not sure if we should introduce that in this single spot.

stable/anchore-engine/templates/enterprise_feeds_configmap.yaml
```diff
@@ -94,54 +94,51 @@ data:
...
-            enabled: {{ default "true" .Values.anchoreEnterpriseFeeds.nvdv2DriverEnabled }}
+            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.nvdv2DriverEnabled | quote) }}
           vulndb:
-            enabled: {{ default "true" .Values.anchoreEnterpriseFeeds.vulndbDriverEnabled }}
-        {{- if .Values.anchoreEnterpriseFeeds.msrcDriverEnabled }}
+            enabled: {{ default "true" (.Values.anchoreEnterpriseFeeds.vulndbDriverEnabled | quote) }}
           msrc:
-            enabled: true
-            api_key: ${ANCHORE_MSRC_KEY}
+            enabled: {{ default "false" (.Values.anchoreEnterpriseFeeds.githubDriverEnabled | quote) }}
```
Vijay-P, line 130: typo: msrcDriverEnabled

Vijay-P (changes requested):
found a typo

stable/anchore-engine/templates/enterprise_feeds_configmap.yaml
Vijay-P: In the previous chart, the values would not work if the booleans were not quoted. With this update, the chart works in either case. However, this check only works if the boolean is not quoted. I think that this may cause issues for customers who upgrade. If we do not change this, we should be very careful in communicating to customers that they should unquote the booleans in their values files.

Vijay-P: After thorough testing, I recommend using `{{- if eq (.Values.anchoreEnterpriseFeeds.githubDriverEnabled | toString) "true"}}`

stable/anchore-engine/templates/enterprise_feeds_configmap.yaml
Btodhunter: good catch!

stable/anchore-engine/templates/enterprise_feeds_configmap.yaml
Btodhunter: Nice find! `toString` is definitely the better solution here. We should put in an issue for updating all `if` statements that are expecting true/false values to follow this condition.

Vijay-P (dismissed):

Vijay-P (approved):

found-it:
Should the msrc api key be removed from the values file? Makes that change more obvious

https://github.com/anchore/anchore-charts/blob/d2c756ce28a9e2b42a08f5a29446a2d2e25f1325/stable/anchore-engine/values.yaml#L738

found-it:
LGTM

zhill (approved):
lgtm

PR #164, opened 2021-08-31
Fix connection strings for db using TLS

Btodhunter:
Fixed an issue with the DB upgrade job, when using TLS, that didn't allow specifying the root cert path due to not handling white space & escape characters properly.

Fixed sslMode=require getting forced to verify-all because the cert path was always specified in the connection string. See https://www.postgresql.org/docs/9.1/libpq-ssl.html

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.14.3
+version: 1.14.4
 appVersion: 0.10.1
 description: Anchore container analysis and policy evaluation engine service
 keywords:
--- a/stable/anchore-engine/README.md
+++ b/stable/anchore-engine/README.md
@@ -561,7 +561,9 @@ anchoreEnterpriseUI:
 
 ## Install using an existing/external PostgreSQL instance
 
-*Note: it is recommended to use an external Postgresql instance for production installs*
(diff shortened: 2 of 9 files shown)


Vijay-P:
Tested behavior:

values.yaml
```
anchoreGlobal:
```

Result:
```
db_connect: "postgresql://${ANCHORE_DB_USER}:${ANCHORE_DB_PASSWORD}@${ANCHORE_DB_HOST}/${ANCHORE_DB_NAME}"
```

values.yaml
```
anchoreGlobal:
  dbConfig:
    ssl: true
```

Result:
```
db_connect: "postgresql://${ANCHORE_DB_USER}:${ANCHORE_DB_PASSWORD}@${ANCHORE_DB_HOST}/${ANCHORE_DB_NAME}?sslmode=verify-full&sslrootcert=/home/anchore/certs/"
```

values.yaml
```
anchoreGlobal:
  dbConfig:
    ssl: true
    sslMode: require
```

Result:
```
db_connect: "postgresql://${ANCHORE_DB_USER}:${ANCHORE_DB_PASSWORD}@${ANCHORE_DB_HOST}/${ANCHORE_DB_NAME}?sslmode=require"
```

Vijay-P (dismissed):

Btodhunter:
Added some more documentation and fixed up the chart version so it doesn't conflict.

Vijay-P (approved):

PR #180, opened 2021-11-09
Honor the timeout seconds config in kai

found-it:
Honor the timeout-seconds config, the way it was set up in the configmap the app would silently fail to read the config option.


Btodhunter (approved):
lgtm!

PR #187, opened 2022-02-11
update the existing secret comment for db secret

found-it:
Updated the comment for the feeds database existing secret to reflect that the ANCHORE_GITHUB_TOKEN is a required field if using the the `existingSecret`.


Btodhunter (approved):
lgtm!

PR #191, opened 2022-03-03
Enable grypedb default

Btodhunter:
Enables the GrypeDB builder within the Enterprise Feeds service by default.

### DIFF
--- a/stable/anchore-engine/README.md
+++ b/stable/anchore-engine/README.md
@@ -244,6 +244,32 @@ See the anchore-engine [CHANGELOG](https://github.com/anchore/anchore-engine/blo
 A Helm post-upgrade hook job will shut down all previously running Anchore services and perform the Anchore database upgrade process using a Kubernetes job. 
 The upgrade will only be considered successful when this job completes successfully. Performing an upgrade will cause the Helm client to block until the upgrade job completes and the new Anchore service pods are started. To view progress of the upgrade process, tail the logs of the upgrade jobs `anchore-engine-upgrade` and `anchore-enterprise-upgrade`. These job resources will be removed upon a successful Helm upgrade.
 
+## Chart version 1.17.0
+
+Chart version 1.17.0 is a tightly scoped release, specifically aimed at enabling the Grype DB Builder within the Anchore Enterprise Feeds service. This upgrade will allow users to easily switch from the legacy vulnerability provider to the next-gen Grype vulnerability provider, without any vulnerability feed downtime. The impacts of this upgrade are as follows:
+
+* For deployments currently utilizing the legacy vulnerability provider, configured with `.Values.anchorePolicyEngine.vulnerabilityProvider=legacy`, this upgrade will enable the GrypeDB Builder feeds source on the Enterprise Feeds service.
+  * Grype is the only supported vulnerability provider for Anchore Enterprise v4.0.0 and higher (coming with chart v1.18.0).
+  * The GrypeDB builder can be manually disabled for legacy deployments using `.Values.anchoreEnterpriseFeeds.grypeDriverEnabled=false`
+
+  ### WARNING
+
+  After this upgrade, the Enterprise Feeds service requires a minimum of 10GB of memory allocated. **Failure to allocate adequate resources to this pod will result in crash loops and an unavailable feeds service.** Resource allocation example:
+
+    ```yaml
+    anchoreEnterpriseFeeds:
+      resources:
+        limits:
+          cpu: 1
+          memory: 10G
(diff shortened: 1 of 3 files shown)


stable/anchore-engine/README.md
```diff
@@ -244,6 +244,32 @@ See the anchore-engine [CHANGELOG](https://github.com/anchore/anchore-engine/blo
 A Helm post-upgrade hook job will shut down all previously running Anchore services and perform the Anchore database upgrade process using a Kubernetes job. 
 The upgrade will only be considered successful when this job completes successfully. Performing an upgrade will cause the Helm client to block until the upgrade job completes and the new Anchore service pods are started. To view progress of the upgrade process, tail the logs of the upgrade jobs `anchore-engine-upgrade` and `anchore-enterprise-upgrade`. These job resources will be removed upon a successful Helm upgrade.
 
+## Chart version 1.17.0
+
+Chart version 1.17.0 is a tightly scoped release, specifically aimed at enabling the Grype DB Builder within the Anchore Enterprise Feeds service. This upgrade will allow users to easily switch from the legacy vulnerability provider to the next-gen Grype vulnerability provider, without any vulnerability feed downtime. The impacts of this upgrade are as follows:
```
zhill, line 249: I think it would help users to lead with "This update impacts Enterprise users only." So all others know there is no change here for them to worry about.

Instead of "next-gen" use "v2" scanner, and instead of "legacy" let's use "v1" when referring to the scanner but not the specific config value, since those values have to be specific.

stable/anchore-engine/README.md
```diff
@@ -244,6 +244,32 @@ See the anchore-engine [CHANGELOG](https://github.com/anchore/anchore-engine/blo
...
+
+* For deployments currently utilizing the legacy vulnerability provider, configured with `.Values.anchorePolicyEngine.vulnerabilityProvider=legacy`, this upgrade will enable the GrypeDB Builder feeds source on the Enterprise Feeds service.
```
zhill, line 251: I'd prefer "GrypeDB driver" instead of "GrypeDB Builder" since that is more descriptive

stable/anchore-engine/README.md
```diff
@@ -244,6 +244,32 @@ See the anchore-engine [CHANGELOG](https://github.com/anchore/anchore-engine/blo
...
+  * Grype is the only supported vulnerability provider for Anchore Enterprise v4.0.0 and higher (coming with chart v1.18.0).
```
zhill, line 252: I don't think this is necessary to talk about a future release. Let's make it clear this change helps any Enterprise customer moving from the v1 scanner to the v2 scanner reduce their downtime, and that is covered in the intro paragraph I think.

stable/anchore-engine/README.md
```diff
@@ -244,6 +244,32 @@ See the anchore-engine [CHANGELOG](https://github.com/anchore/anchore-engine/blo
...
+  * The GrypeDB builder can be manually disabled for legacy deployments using `.Values.anchoreEnterpriseFeeds.grypeDriverEnabled=false`
```
zhill, line 253: "GrypeDB driver"

stable/anchore-engine/README.md
zhill, line 249:
```suggestion
Chart version 1.17.0 is a purely Enterprise focused release. Anchore Engine users will see no change in behavior from this release.

For Enterprise users, this release specifically helps reduce downtime needed during the transition from the v1 scanner to the v2 scanner. This version sets the GrypeDB driver to run in the feed service v1-scanner deployments so that the GrypeDB is ready when the update to the v2 scanner is made and thus reduces effective downtime during the maintenance window needed for that configuration change.

It is recommended that users upgrade to this chart version prior to changing the scanner configuration to move from the V1 scanner to the V2 scanner.
```
Then insert the resource warning here before going into the bullet points about config options.

wehand55: I agree with Zach's comments, I'd add a slight change, you can use "v1 (legacy) scanner" for this helm release.

stable/anchore-engine/README.md
wehand55: I like the use of GrypeDB driver, instead of Builder, but I don't recommend making a change in 1.17.  We need to make a change in Enterprise docs etc.   Let's plan to change terms in later helm versions?

zhill: I'm only commenting on this specific description text for the README. Agree the chart content itself is a different discussion and should not be changed for 1.17

zhill (approved):
lgtm

PR #194, opened 2022-03-15
Update to Enterprise v4.0.0

Btodhunter:


zhill (approved):
lgtm. I did spot test of defaults with new 4.0 images, but did not test all conditions or cases. But things look good.

PR #199, opened 2022-03-30
adding default value for catalog imageGCMaxWorkerThreads

d3vdr3y:
Signed-off-by: Andre Neufville <leoneuf1@gmail.com>

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.18.1
+version: 1.18.2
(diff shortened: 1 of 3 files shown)


blang9238 (dismissed):
lgtm!

Btodhunter (changes requested):
I talked with Vijay about this configuration option, and it is not something we want to expose for customers to change at will. They should get guidance from CX for making this change. To accommodate that, we should remove the default value from the values file `anchoreCatalog.imageGCMaxMaxWorkerThreads`

Btodhunter (approved):

Btodhunter (approved):

PR #200, opened 2022-03-31
Check for actual ingress object in api version checks

found-it:
On older versions of kubernetes the `networking.k8s.io/v1` API exists however the ingress object is not present. Simply checking for the existence of the API will cause the conditional to incorrectly set the `apiVersion`. Helm has the capability to check for the object itself so this PR just updates the check for the Ingress object.

https://helm.sh/docs/chart_template_guide/builtin_objects/#helm

> Capabilities.APIVersions.Has $version indicates whether a version (e.g., batch/v1) or resource (e.g., apps/v1/Deployment) is available on the cluster.

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.18.2
+version: 1.18.1
 appVersion: 1.1.0
 description: Anchore container analysis and policy evaluation engine service
 keywords:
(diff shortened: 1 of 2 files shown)


Btodhunter:
@found-it this LGTM! Is it still a WIP?

found-it:
It can be merged, I was testing on an older version of k3s and ran into some weirdness with the way helm understood the ingress object vs k3s which is why I still have it as WIP. It's not an issue with this change though, I think it's a weird interaction between k3s and helm, I just haven't had time to get back to it.

found-it:
Okay tested on 1.18, 1.19, and 1.23 and it selected the correct API versions!

PR #203, opened 2022-04-20
Added annotations to metadata for deployments, global and service specific

blang9238:

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.18.4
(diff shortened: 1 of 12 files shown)


Btodhunter:
Thanks for this PR @blang9238! This is a much needed change. One observation I have is that `metadata` is used for all types of annotations, so I think that using `<component>.metadata.annotations` for the config is kind of confusing. Maybe we use `<component>.deploymentAnnotations`? That will keep the values file a bit more flat and better describe what that config is used for.

blang9238:
Yep that definitely makes sense @Btodhunter, I'll make those changes.

Btodhunter (approved):
lgtm!

Issue #206, opened 2022-05-03
Remove kubernetes.io/ingress.class conflicts with ingressClassName


calvinbui:
There is a [default annotation value for kubernetes.io/ingress.class: nginx](https://github.com/anchore/anchore-charts/blob/d9fba112523f8a724b4d28354aec2246ee777ed5/stable/anchore-engine/values.yaml#L88) which is conflicting with `ingressClassName`

```
Ingress.extensions "anchore-engine-anchore-engine" is invalid: annotations.kubernetes.io/ingress.class: Invalid value: "nginx": can not be set when the class field is also set
```

PR #207, opened 2022-05-04
fix: ingress class shouldn't be set by default

sfroment:
Ingress class shouldn't be set by default, and if we were to enable one by default we should enable ingressClassName since the annotations is deprecated

Signed-off-by: sfroment <sfroment42@gmail.com>


calvinbui:
closes #206

Btodhunter:
@calvinbui Thanks for the update! Sorry for the delayed response, LGTM!

Btodhunter (approved):

Btodhunter (approved):

PR #213, opened 2022-07-19
Update Ingress.yaml : Fix APIVersions 

Kilz78:
fix APIVersions


Btodhunter:
Closing as unnecessary. We specifically check for `/Ingress` because we had a customer who had `networking.k8s.io/v1` available on their cluster but it didn't have `ingress` and it broke their configuration.

PR #217, opened 2022-08-22
Add scratch volume to catalog

found-it:
Currently there is no way to mount the catalog scratch space as a volume. This means in read-only file systems we can't configure the scratch space to be writeable. This PR just adds the scratch volume config to the catalog deployment.


hunguyen1 (approved):
lgtm

PR #218, opened 2022-08-25
Dev sync pr

hunguyen1:
Removing vulndb feed driver configurations and adding new anchore_match_exclusion feed. Also bumping image version and Chart verison.


zhill (approved):
lgtm

PR #219, opened 2022-08-29
Switch KAI to XDG config

found-it:
When KAI runs on open shift clusters it will assign a random user. The
underlying viper code will fail when looking in /home/nonroot so the
config file should exist in $XDG_CONFIG_DIRS so the container can be run
as any user and still function properly


hunguyen1 (approved):
lgtm!

PR #220, opened 2022-09-13
provide option for overriding grypeProviderFeedsxternalURL

hunguyen1:
Signed-off-by: Hung Nguyen <hung.nguyen@anchore.com>


found-it:
Added some logic in so it will reference a boolean under the policy engine to point the policy engine upstream while not affecting the normal enterprise feed sync. Left the config out of the values.yaml file since it's "use at your own risk" feature

found-it:
@hunguyen1 if you want to review my commits I think we're good to merge

hunguyen1:
> @hunguyen1 if you want to review my commits I think we're good to merge

@found-it lgtm!

found-it (approved):
lgtm

PR #228, opened 2022-12-07
Admission controller v0.4.0

zhill:
Adds compatibility for k8s 1.22+ but also adds a kubeVersion constraint to the chart of 1.19+ to ensure old k8s versions don't fail trying to deploy this update.

For k8s clusters < 1.19, they should stick with the 0.3.x chart.


zhill:
I expect to rebase/squash this before a merge. Ongoing dev requires both controller code and chart updates so this will be a bit fluid until its ready then will get rebased to a single update for the final review & test.

zhill:
I will re-base this on main after the version updates for tests have landed there.

zhill:
Note, I will update this with the full release tag now that we've ironed out the testing kinks.

Btodhunter (approved):
lgtm

PR #229, opened 2022-12-07
chore: update k8s test matrix for chart updates

zhill:
Signed-off-by: Zach Hill <zach@anchore.com>

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kubernetesVersion: ["v1.14.10", "v1.22.0"]
+        kubernetesVersion: ["v1.19.16", "v1.22.0", "v1.25.0"]
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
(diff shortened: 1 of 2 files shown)


Btodhunter (dismissed):
lgtm

.github/workflows/test.yaml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        kubernetesVersion: ["v1.14.10", "v1.22.0"]
+        kubernetesVersion: ["v1.19.0", "v1.22.0", "v1.25.0"]
```
Btodhunter, line 11: I've run into issues with older versions of the `kindest:node` image, the testing for v1.19.0 might pass if you use the latest kindest image `v1.19.16`

zhill: I'm trying 1.21 instead which is the oldest version EKS supports. You ok with 1.21, 1.22, 1.25?

zhill: nvm, there is no 1.21.0 kindest/node image so I'm using 1.19.16 anyway.

Btodhunter:
lgtm

Btodhunter (approved):

PR #154, opened 2021-08-12
Add option to use Cloud SQL IAM authentication

rogeriobastos:
CloudSQL supports IAM database authentication which allows a passwordless access to database instance using service account key. To use this feature with cloud_sql_proxy, we need to pass `-enable_iam_login` parameter.

References:
* https://cloud.google.com/sql/docs/postgres/authentication?hl=en-us
* https://github.com/GoogleCloudPlatform/cloudsql-proxy#-enable_iam_login

PR #232, opened 2022-12-22
cloudsql fixes & stale PR implementations

Btodhunter:
Allows passing a list of custom arguments to the cloudsql container entrypoint. This implements fixes that address the same issues as the following stale PRs: #154 & #155


HN23 (approved):
lgtm

PR #239, opened 2023-01-12
Anchore admission controller and EKS

ndegory:
The kubeVersion field of the anchore-admission-controller helm chart v0.4.3 fails to validate the Kubernetes version when deploying on EKS.

```
chart requires kubeVersion: ^1.19.0 which is incompatible with Kubernetes v1.23.13-eks-fb459a0
```

This is a known issue with EKS using the pre-release syntax and breaking the semVer comparison. Changing the control to ^1.19.0-0 fixes the deployment on EKS.

This PR also contains the addition of an optional imagePullsecrets value, and moves the hardcoded init-ca image to the values.yaml.


stable/anchore-admission-controller/values.yaml
```diff
@@ -103,6 +107,9 @@ credentials: {}
 
 # Settings related to init-ca pod
 initCa:
+  registry: ""
```
HN23, line 110: Nit: To help prevent confusion for folks who aren't inclined to dig deep into the templates, should we populate initCa.registry=docker.io in the values file rather than using the default function in the template? alternatively, we could keep the current convention of having a flat `image` key.

HN23 (dismissed):

stable/anchore-admission-controller/values.yaml
ndegory: sure, will change that

HN23 (approved):
lgtm! thanks for this contribution @ndegory !

authentichz:
Thank you @ndegory!

When this change can be merged?

PR #242, opened 2023-01-20
adding labels and resources template

Kaizhe:
- template labels
- template resources

### DIFF
--- a/stable/anchore-admission-controller/templates/deployment.yaml
+++ b/stable/anchore-admission-controller/templates/deployment.yaml
@@ -3,6 +3,9 @@ kind: Deployment
 metadata:
   name: {{ template "anchore-admission-controller.fullname" . }}
   labels: {{- include "anchore-admission-controller.labels" . | nindent 4 }}
+  {{- with .Values.labels}}
+    {{ toYaml . | nindent 4 }}
+  {{- end }}
   annotations:
   {{- with .Values.annotations }}
     {{ toYaml . | nindent 4 }}
@@ -16,6 +19,9 @@ spec:
   template:
     metadata:
       labels: {{- include "anchore-admission-controller.labels" . | nindent 8 }}
+      {{- with .Values.labels}}
+        {{ toYaml . | nindent 8 }}
+      {{- end }}
       annotations:
       {{- with .Values.annotations }}
         {{ toYaml . | nindent 8 }}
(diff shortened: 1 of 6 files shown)


zhill:
@Kaizhe thanks for the PR! To get it to pass the linter, can you add a patch version bump to the chart version? I think 0.4.5 is the right version for this change. Thanks!

Kaizhe:
> @Kaizhe thanks for the PR! To get it to pass the linter, can you add a patch version bump to the chart version? I think 0.4.5 is the right version for this change. Thanks!

Done, thanks!

zhill:
I can help with getting and keeping CI moving, but others will probably need to do full review/approval.

stable/anchore-admission-controller/values.yaml
```diff
@@ -93,6 +93,7 @@ tolerations: []
 ## Constrain which nodes your pod is eligible to be scheduled on, based on labels on the node
 ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
 affinity: {}
+labels: {}
```
Btodhunter, line 96: Should we rename this `extraLabels` to indicate that it's additive to the labels that currently exist in the helper template?

stable/anchore-admission-controller/templates/deployment.yaml
```diff
@@ -3,6 +3,9 @@ kind: Deployment
 metadata:
   name: {{ template "anchore-admission-controller.fullname" . }}
   labels: {{- include "anchore-admission-controller.labels" . | nindent 4 }}
+  {{- with .Values.labels}}
```
Btodhunter, line 6: We should include these `with` statements in the `anchore-admission-controller.labels` helper template rather than sticking them into each resource template directly. This will help keep our templates DRY.

Btodhunter (changes requested):
Thanks for the PR @kaizhe! These are great improvements to the chart. Just a small change request to keep the templates following the same conventions we use throughout our charts. All merge as soon as the change request is addressed!

stable/anchore-admission-controller/values.yaml
Kaizhe: Done. thanks!

Btodhunter (approved):
lgtm!

PR #243, opened 2023-01-23
chore: refactor CI workflow to move linter to its own flow

zhill:
Adds new linter workflow and removes that step from the test workflow.
Should make matrix tests easier and linting issues faster to resolve.

Signed-off-by: Zach Hill <zach@anchore.com>


Btodhunter (approved):
lgtm!

PR #244, opened 2023-01-24
Add existing secret check

Btodhunter:
This will prevent users who have `anchoreGlobal.existingSecrets` set from a failed upgrade if they didn't update their values file and secrets for chart v1.21.0


HN23 (approved):
looks good to me!

PR #252, opened 2023-02-24
Fix dryrun upgrade

Btodhunter:
Tested with the following:
* Confirmed that upgrades from Enterprise v4.1.0 showed the warning about upgrade support during dry-run upgrades & failed during non dry-runs
```
helm install anchore anchore/anchore-engine --version 1.18.0 --set anchoreEnterpriseGlobal.enabled=true
helm upgrade anchore . --set anchoreEnterpriseGlobal.enabled=true --dry-run
helm upgrade anchore . --set anchoreEnterpriseGlobal.enabled=true 
```
* Confirmed that upgrading from Enterprise v4.2.0 was successful
```
 helm install anchore anchore/anchore-engine --set anchoreEnterpriseGlobal.enabled=true --version 1.20.0
 helm upgrade anchore . --set anchoreEnterpriseGlobal.enabled=true 
```

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.22.3
+version: 1.22.4
 appVersion: 1.1.0
 description: Anchore container analysis and policy evaluation engine service
 keywords:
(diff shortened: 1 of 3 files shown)


HN23 (approved):
lgtm!

PR #254, opened 2023-03-21
[anchore-admission-controller] annotations for the validating webhook configuration

ndegory:
Add configurable annotations for the validating webhook configuration.
This can be used for instance to tell cert-manager to inject a CA cert instead of passing it in the values file.


stable/anchore-admission-controller/templates/webhook.yaml
```diff
@@ -4,6 +4,9 @@ kind: ValidatingWebhookConfiguration
 metadata:
   name: {{ template "anchore-admission-controller.name" . }}-admission.anchore.io
   labels: {{- include "anchore-admission-controller.labels" . | nindent 4 }}
+  {{- with .Values.apiService.webhook.annotations }}
```
zhill, line 7: It looks like this needs to be '.Values.apiService.annotations' based on the annotation location in the values.yaml and that the 'webhook' level is for the webhook config rather than the api service. Can you verify if that is what's expected?

PR #255, opened 2023-03-24
fix certifi cacert issues when users are supplying custom certificates

Btodhunter:
Added venv activation to `anchoreGlobal.doSourceAtEntry.enabled=true`. This will cause the certify cacert to be copied properly when the docker-entrypoint is run. Provides a temporary solution for users running into TLS issues on Anchore Enterprise v4.5.0 when providing custom certificates with `anchoreGlobal.certStoreSecretName`


svietry (approved):
LGTM

PR #260, opened 2023-04-04
Add Container securityContext

chrisad2:
Adds securityContext at the container level and a new value in `anchoreGlobal` to set this context. I have been maintaining this in my own environment and rebasing with every new chart release, but I'd like to get this upstream.

- I can't test the cloudsql proxy changes (I don't use cloudsql)
- I'm unsure in the upgrade jobs whether this is the best way to organize this change


Btodhunter (approved):
I tested this change with cloudsql and it worked great! I think the logic used in the upgrade jobs is also good to go. Thanks for the improvement!

PR #262, opened 2023-04-20
Add quotes to db config environment vars

Btodhunter:
To prevent '!' from being interpreted by YAML we need to quote out all env vars used for db config variables.


HN23 (approved):
lgtm

PR #263, opened 2023-04-21
Add Labels from Values Files to Upgrade Jobs

chrisad2:
Extends `.Values.anchoreGlobal.labels` to apply to all three upgrade jobs.

Additionally adds ability to specify labels _per job_ in the same pattern as other templates (such as the deployment templates).


Btodhunter (dismissed):
Thanks for the addition @chrisad2! LGTM!

PR #271, opened 2023-05-30
feat: updates mountPath for RedHat sourced Postgres image when using OpenShift

zhill:
The RedHat postgres images use a different data path and ignore PGDATA environment vars, so the mount path must be updated in order to ensure the DB PVC is used correctly for data.

Without this update, the DB will work, but DB data is lost when the pod is deleted.

### DIFF
--- a/stable/anchore-engine/Chart.yaml
+++ b/stable/anchore-engine/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.26.4
+version: 1.26.5
 appVersion: 1.1.0
 description: Anchore container analysis and policy evaluation engine service
 keywords:
--- a/stable/anchore-engine/README.md
+++ b/stable/anchore-engine/README.md
@@ -127,7 +127,7 @@ postgresql:
   - name: POSTGRESQL_USER
     value: anchoreengine
   - name: POSTGRESQL_PASSWORD
-    value: anchore-postgres,123
+    value: <PGPASSWORD>
   - name: POSTGRESQL_DATABASE
     value: anchore
   - name: PGUSER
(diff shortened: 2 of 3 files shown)


HN23 (approved):

zhill:
@HN23 fixed a typo in the comment that was bugging me, so need a re-review :(

Btodhunter (approved):
lgtm

zhill:
I need to update the README as well since we document the changes for OpenShift there.

stable/anchore-engine/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: anchore-engine
-version: 1.25.1
+version: 1.25.2
```
HN23, line 3: we should sync this with main + 1

HN23:
Also, i've confirmed that having the mountPath for persisting data between pod restarts works in an openshift deployment. It should be noted though that upgrading an existing deployment without the mountPath to setting the mountPath will still see the data loss. 

ie. `helm upgrade anchore --set postgresql.persistence.mountPath=/var/lib/pgsql/data` will not preserve the data from the current installation yet, so be sure to back up any data first.

HN23:
Closing PR. Will be added to enterprise chart instead with docs for upgrading.

zhill:
> Also, i've confirmed that having the mountPath for persisting data between pod restarts works in an openshift deployment. It should be noted though that upgrading an existing deployment without the mountPath to setting the mountPath will still see the data loss (from 13 to 13).
> 
> ie. `helm upgrade anchore --set postgresql.persistence.mountPath=/var/lib/pgsql/data` will not preserve the data from the current installation yet, so be sure to back up any data first.

In this case you mean if the user wants to switch from the regular PG image to the RedHat PG image but use the same volume? Then yes, agree that is not supported because the images are incompatible in how they lay out the PG data in the FS. Just making sure we're describing the same use-case and not a new error condition.

HN23 (approved):
lgtm

PR #273, opened 2023-06-15
adding upgrade job via preupgrade hook in helm

HN23:
enabling the preupgrade hook will create a service account so that the job ran as part of the preupgrade hook can scale down the deployments before doing the database migrations.

This is disabled by default and so no changes should be seen by the user unless they update anchoreGlobal.usePreupgradeHook=true in which case, it will no longer be ran as previous, but be ran in the preupgrade hook job instead.


stable/anchore-engine/templates/anchore_upgrade_role.yaml
```diff
@@ -0,0 +1,41 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app: {{ template "anchore-engine.fullname" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+    {{- with .Values.anchoreGlobal.labels }}
+    {{ toYaml . | nindent 4 }}
+    {{- end }}
+  {{- if .Values.anchoreGlobal.annotations }}
```
Btodhunter, line 16: I think all these templates should probably have the labels/annotations/etc from  values.[service]upgradeJob as well as globals.

stable/anchore-engine/templates/anchore_upgrade_role.yaml
```diff
@@ -0,0 +1,41 @@
...
+  annotations:
+    {{- with .anchoreGlobal.annotations }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+  {{- end }}
+rules:
```
Btodhunter, line 22: I'm assuming this is the minimum set of permissions needed for `kubectl scale`?

stable/anchore-engine/values.yaml
```diff
@@ -308,6 +308,10 @@ anchoreGlobal:
       failureThreshold: 3
       successThreshold: 1
 
+  # using the preupgrade hook will create a service account that will use
```
Btodhunter, line 311: I would add a note here that this is useful for deployments using `helm upgrade --wait` or ArgoCD. Also mention that it create the role/rolebinding as well as the SA, maybe even mention the permissions granted to that SA as well.

blang9238: +1 on adding the note about `helm upgrade --wait`

stable/anchore-engine/templates/anchore_upgrade_role.yaml
HN23: added

stable/anchore-engine/templates/anchore_upgrade_role.yaml
HN23: correct. I also added `pods` so that a second initContainer can watch to make sure all the pods are actually down before completing.

stable/anchore-engine/values.yaml
HN23: added

HN23:
on a version lower than 1.24.0
```
helm upgrade puppy -f puppy.yaml . --wait
Error: UPGRADE FAILED: execution error at (anchore-engine/templates/_helpers.tpl:461:12): using the preupgrade hook for upgrades requires the chart to be on at least version 1.24.0. Please upgrade to 1.24.0 or newer before using the preupgrade hook.
```

After upgrading to 1.24.0 first, it works as expected:
```
helm upgrade puppy anchore/anchore-engine --version=1.24.0 -f puppy.yaml
helm upgrade puppy -f puppy.yaml . --wait
```

```
uppy-pre-upgrade-job-g7hzk confirm-service-scaled-down waiting for pods to go down
puppy-pre-upgrade-job-g7hzk confirm-service-scaled-down waiting for pods to go down
puppy-pre-upgrade-job-g7hzk confirm-service-scaled-down waiting for pods to go down
puppy-pre-upgrade-job-g7hzk confirm-service-scaled-down waiting for pods to go down
puppy-pre-upgrade-job-g7hzk upgrade-enterprise-db [MainThread] 2023-06-17T03:55:03.926650 [anchore_enterprise.db.upgrade.upgrade/run_upgrade()] [INFO] Performing upgrade tasks to upgrade from 22 to 24
puppy-pre-upgrade-job-g7hzk upgrade-enterprise-db [MainThread] 2023-06-17T03:55:03.927806 [anchore_enterprise.db.upgrade.upgrade/do_upgrade()] [INFO] Upgrading DB: from=22 to=24
```

Btodhunter (approved):
lgtm!

Btodhunter:
Successfully tested the following upgrade paths:

v1.22.0 -> 1.26.1 w/ preupgrade hook
v1.22.0 -> 1.26.1
v1.24.0 -> 1.26.1 w/ preupgrade hook
v1.24.0 -> 1.26.1

PR #277, opened 2023-07-18
update upgrade jobs to use variables instead of lookup

HN23:


Btodhunter (commented):
We should delete the `postgres.anchore-db.port` helper if it's no longer needed.

HN23:
> We should delete the `postgres.anchore-db.port` helper if it's no longer needed.

removed that as well as the feeds-db one.

Btodhunter (approved):
lgtm

Btodhunter:
Tested the following:

```
using existing secrets going from 1.23.0 -> 1.26.6
no existing secrets 1.23.0 -> 1.26.6
using existing secrets going from 1.25.0 -> 1.26.6
no existing secrets 1.25.0 -> 1.26.6
```

PR #278, opened 2023-08-21
feat: bump to latest k8s-inventory release

bradleyjones:
Signed-off-by: Bradley Jones <bradley.jones@anchore.com>

### DIFF
--- a/stable/k8s-inventory/Chart.yaml
+++ b/stable/k8s-inventory/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: k8s-inventory
-version: 0.0.2
+version: 0.0.4
 appVersion: "1.0.0"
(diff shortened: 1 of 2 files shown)


stable/k8s-inventory/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: k8s-inventory
-version: 0.0.2
+version: 0.0.4
```
bradleyjones, line 3: @anchore/devops I've bumped to 0.0.4 so this chart is inline with anchore-charts-dev so the rebase is clean but if you want this to go back to 0.0.3 and we clean up anchore-charts-dev that's fine too just let me know

HN23:
We are going to close this and merge over from charts-dev since this change is already in.

PR #291, opened 2023-09-13
adding check for readme generator

HN23:
This is so a PR going in with both engine and enterprise changes will not fail as engine does not have the values annotations.


Btodhunter (approved):
lgtm

PR #310, opened 2023-10-28
Enterprise v1.0.0

Btodhunter:


stable/enterprise/README.md
```diff
@@ -568,213 +669,226 @@ ui:
...
-    helm install ${ENTERPRISE_RELEASE} -n ${NAMESPACE} -f ${VALUES_FILE_NAME} --set upgradeJob.force=true anchore/enterprise
-    ```
+    export NAMESPACE=anchore
+    export ENTERPRISE_RELEASE=<YOUR_ENTERPRISE_RELEASE_NAME>
+    export ENTERPRISE_VALUES_FILE=${PWD}/output/enterprise.my-values-file.yaml
 
-    > **Note:** The `upgradeJob.force` flag is required to force the upgrade job to run upon installation. This value is not needed for future upgrades. Remember to unset it if passing it in via the command line or helm may persist the value.
+    # If you are not using existing secrets, uncomment the following export commands
+    #
+    # export ADMIN_PASS=$(kubectl get secret -n ${NAMESPACE} ${ENGINE_RELEASE}-anchore-engine-admin-pass -o jsonpath="{.data.ANCHORE_ADMIN_PASSWORD}" | base64 -d -)
```
HN23, line 753: the `base64 -d -` caused me to get an "invalid argument -" on my mac. taking off the `-` gave the decoded pass

HN23 (dismissed):

stable/enterprise/values.yaml
```diff
@@ -1061,7 +1061,7 @@ rbacManager:
     annotations: {}
     labels: {}
     nodePort: ""
-    apiVersion: v2
+    apiVersion: v1
```
HN23, line 1064: we should update the feeds values.yaml to be v1 as well.

stable/enterprise/README.md
Btodhunter: The `-` is needed for linux but macos doesn't need it. I'm just going to leave it because the error is pretty obvious.

stable/enterprise/values.yaml
Btodhunter: done

HN23 (approved):

PR #311, opened 2023-11-01
Minor Enterprise v1.0.0 bugfixes

Btodhunter:
* Removed the hardcoded `v2` API configurations from the UI
* Updated documentation with deployment differences from `anchore-engine`


HN23 (dismissed):

.github/workflows/test.yaml
```diff
@@ -94,4 +94,4 @@ jobs:
 
     - name: Run chart-testing
       if: steps.list-changed.outputs.CHANGED == 'true'
-      run: ct install --config ct-config.yaml
+      run: ct install --config ct-config.yaml --helm-extra-args "--timeout 600s"
```
HN23, line 97: need to take timeout out of helm-extra-args for now because a bug in ct-testing is passing this to lint, which helm lint doesn't have a --timeout flag.
https://github.com/helm/chart-testing/pull/605/files
when this is merged, we can put the helm-extra-args back in the ct-testing config file.

HN23 (dismissed):

Issue #324, opened 2023-12-18
`anchoreConfig.analyzer.max_threads` still appears in 5.x charts, despite no longer being supported


0xDEC0DE:
According to https://docs.anchore.com/current/docs/releasenotes/500/#deprecations , multiple analyzer threads are no longer supported, but they are still documented, and a migration path is defined in the charts.

If they are no longer supported, they should be removed.

PR #325, opened 2023-12-18
Remove anchoreConfig.analyzer.max_threads from 5.x charts

0xDEC0DE:
According to the docs, this feature is no longer supported; remove it.

Fixes: Issue #324


stable/anchore-engine/templates/engine_configmap.yaml
```diff
@@ -165,7 +165,6 @@ data:
         cycle_timer_seconds: 1
         cycle_timers:
           {{- toYaml .Values.anchoreAnalyzer.cycleTimers | nindent 10 }}
-        max_threads: {{ .Values.anchoreAnalyzer.concurrentTasksPerWorker }}
```
0xDEC0DE, line 168: Arguably this should be left alone, since it's nominally for configuring 4.x, but ¯\\\_(ツ)_/¯

Btodhunter: I'm going to put this back, I would like to keep `anchore-engine` 4.x friendly.

0xDEC0DE: Reverted.

Btodhunter:
@0xDEC0DE You're the best! Thanks for the quick update, you beat me to it.

Btodhunter:
For unit tests you'll want to run `helm unittest stable/enterprise -u` to update the snapshots.

PR #328, opened 2023-12-22
making kubectl scale down preupgrade container configurable

HN23:
Allow for changing the initContainers image in the preupgrade job that does kubectl scale down


Btodhunter (approved):
lgtm

PR #329, opened 2023-12-22
making kubectl scale down preupgrade container configurable

HN23:
Allow for changing the initContainers image in the preupgrade job that does kubectl scale down


Btodhunter (dismissed):
lgtm

Btodhunter:
@HN23 We should merge the update to the feeds chart and then bump the feeds dep in this PR as well.

Btodhunter (approved):

PR #338, opened 2024-02-02
stable/enterprise: Update Anchore Enterprise to v5.2.0

Btodhunter:


stable/enterprise/templates/envvars_configmap.yaml
```diff
@@ -50,7 +50,7 @@ data:
   ANCHORE_ENTERPRISE_REPORTS_VULNERABILITIES_BY_K8S_CONTAINER: "true"
   ANCHORE_ENTERPRISE_REPORTS_VULNERABILITIES_BY_K8S_NAMESPACE: "true"
 {{- if eq (toString .Values.anchoreConfig.catalog.runtime_inventory.image_ttl_days) "-1" }}
-  {{- fail "The Value `-1` is no longer valid for `.Values.anchoreConfig.catalog.runtime_inventory.image_ttl_days`. Please use `.Values.anchoreConfig.catalog.runtime_inventory.image_ingest_overwrite=true` to force runtime inventory to be overwritten upon every update for that reported context. `.Values.anchoreConfig.catalog.runtime_inventory.image_ttl_days` must be set to a value >1." -}}
+  {{- fail "The Value `-1` is no longer valid for `.Values.anchoreConfig.catalog.runtime_inventory.image_ttl_days`. Please use `.Values.anchoreConfig.catalog.runtime_inventory.image_ingest_overwrite=true` to force runtime inventory to be overwritten upon every update for that reported context. `.Values.anchoreConfig.catalog.runtime_inventory.inventory_ttl_days` must be set to a value >1." -}}
```
svietry, line 53: image_ingest_overwrite=true

This should be 

inventory_ingest_overwrite=true

svietry, line 53:
```suggestion
  {{- fail "The Value `-1` is no longer valid for `.Values.anchoreConfig.catalog.runtime_inventory.image_ttl_days`. Please use `.Values.anchoreConfig.catalog.runtime_inventory.inventory_ingest_overwrite=true` to force runtime inventory to be overwritten upon every update for that reported context. `.Values.anchoreConfig.catalog.runtime_inventory.inventory_ttl_days` must be set to a value >1." -}}
```

svietry: maybe we should fail altogether when image_ttl_days or image_ingest_overwrite is set at all.

stable/enterprise/values.yaml
```diff
@@ -532,6 +535,19 @@ anchoreConfig:
     ## The value should be a number followed by "w", "d", or "h" to represent weeks, days or hours
     async_execution_timeout: 48h
 
+    ## @param anchoreConfig.reports.cycle_timers.reports_scheduled_queries Interval  in seconds to check for scheduled queries that need to be run
+    ##
+    cycle_timers:
+      reports_scheduled_queries: 600
+
+    ## @param anchoreConfig.reports.use_volume Configure the reports worker to buffer report generation to disk instead of in memory
```
bot190, line 543:
```suggestion
    ## @param anchoreConfig.reports.use_volume Configure the reports service to buffer report generation to disk instead of in memory
```

stable/enterprise/values.yaml
```diff
@@ -1011,8 +1111,15 @@ reports:
   extraEnv: []
 
   ## @param reports.resources Resource requests and limits for Anchore Reports pods
+  ## ref: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
+  ## Commented values below are just a suggested baseline. Contact Anchore support for deployment specific recommendations.
```
bot190, line 1115: It may be worth putting a note here pointing to the `use_volume` config option as that will impact memory requirements. Otherwise they'll need to set higher memory limits here if they have large amounts of data.

stable/enterprise/README.md
```diff
@@ -1369,6 +1382,15 @@ For the latest updates and features in Anchore Enterprise, see the official [Rel
...
+  - :warning: **WARNING:** Values file changes necessary:
+    - If you are using a custom port for the reports service, previously set with `api.service.reportsPort`, you will need to update your values file to use `reports.service.port` instead.
+    - Resource requests & limits were previously set for both reports pods found in the `reports_deployment` and `api_deployment` using the `reports.resources` section of the values file. These have been split into separate deployments and the resources are now set in the `reports.resources` and `reports_worker.resources` sections of the values file. If you are using custom resources, you will need to update your values file to reflect this change.
+- The reports service no longer has an accessible API endpoint, all API requests should be made to the API service. This version of the chart removed deprecated ingress configurations to accommodate this change. Update your values file to remove all references to the `reports` service in the `ingress` section.
```
zhill, line 1392: "The reports service no longer has an accessible API endpoint..." I think it would be a bit more clear with:

"The reports service is now an internal service and the GraphQLAPI/ReportsAPI is served to users by the API service and routed internally in the deployment as needed."

HN23 (approved):

PR #339, opened 2024-02-02
fix: updates handling of github token to use a default string instead of null

zhill:
If a user does not set the github token value in the feeds chart, the env var should be an empty string rather than "null"
because that causes errors in the configuration validation later when the feeds pod starts.


asomya (dismissed):

stable/feeds/tests/secret_test.yaml
```diff
@@ -0,0 +1,26 @@
...
+templates:
+  - secret.yaml
+release:
+  name: test-release
+  namespace: test-namespace
+chart:
+  version: 9.9.9
+  appVersion: 9.9.9
+tests:
+  - it: should render github driver token as empty string when not set
```
anchoreci, line 11: just to be clear, this doesnt actually set the driver token as an empty string (ie. ANCHORE_GITHUB_TOKEN=""). It doesnt set it at all in the secret, and so it'll use the env_vars value of ANCHORE_GITHUB_TOKEN="default-unset"

HN23, line 11: just to be clear, this doesnt actually set the driver token as an empty string (ie. ANCHORE_GITHUB_TOKEN=""). It doesnt set it at all in the secret, and so it'll use the env_vars value of ANCHORE_GITHUB_TOKEN="default-unset"

zhill: correct. so the description needs an update in the test. good catch!

That was already the existing behavior and my change just sets the default now to 'default-unset' in the ENV always, so that will be overridden only when the user enables driver and then it will put the token value into the secret.

I may look at guarding against the combined values of the driver enabled but no token set (unless there is an external secret mounted), but that seems like a follow-up to this PR.

Btodhunter (approved):

stable/feeds/tests/secret_test.yaml
zhill: thanks!

HN23 (approved):

PR #345, opened 2024-02-12
Add component values to feeds upgrade job

chrisad2:
The feeds upgrade job did not seem to honor feedsUpgradeJob.labels

Copied the pattern from the enterprise chart
- added a component name
- added the component merge to calls to enterprise.common.*

In my particular use case, I needed to disable a sidecar by adding a label to the upgrade job.

### DIFF
--- a/stable/feeds/templates/hooks/post-upgrade/upgrade_job.yaml
+++ b/stable/feeds/templates/hooks/post-upgrade/upgrade_job.yaml
@@ -1,12 +1,13 @@
 {{- if and .Values.feedsUpgradeJob.enabled .Values.feedsUpgradeJob.usePostUpgradeHook -}}
+{{- $component := "upgradeJob" -}}
 
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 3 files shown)


Btodhunter:
@chrisad2 Unfortunately the labels/annotations/environment common helper templates in the feeds chart cannot take in a component argument. I'll try to find time by the end of the week to add a commit to this PR to add that functionality to those helper templates.

stable/feeds/templates/hooks/post-upgrade/upgrade_job.yaml
```diff
@@ -1,12 +1,13 @@
 {{- if and .Values.feedsUpgradeJob.enabled .Values.feedsUpgradeJob.usePostUpgradeHook -}}
+{{- $component := "upgradeJob" -}}
```
Btodhunter, line 2: we'll want this to match the key in values file `feedsUpgradeJob`

Btodhunter (changes requested):

PR #347, opened 2024-02-14
docs: missing namespace in get secret command

bradleyjones:
Signed-off-by: Bradley Jones <bradley.jones@anchore.com>


Btodhunter (approved):
Thanks @bradleyjones! LGTM

PR #358, opened 2024-04-09
Common nodeSelector, affinity, tolerations

chrisad2:
I'd like to add common nodeSelector, affinity, and tolerations to both charts to make values files more DRY.

In my case (and I suspect others), I use the same nodeSelector and tolerations for every service. The intent of this PR is (for both feeds and enterprise charts), to allow setting of those values at the root of the values files for each chart. The changes:

- Use the $components nodeSelector/affinity/tolerations if set
- If not set at $component level, use the common setting

This way I can set these values one time in each chart and not have to maintain it all over my values files (such as in the case of new deployments being created)

### DIFF
--- a/stable/enterprise/Chart.yaml
+++ b/stable/enterprise/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: enterprise
-version: "2.5.0"
+version: "2.5.1"
 appVersion: "5.4.0"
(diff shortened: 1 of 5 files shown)


HN23:
I tested this in my environment using a top level nodeselector, tolerations, and affinity, as well as a service level nodeSelector, tolerations, and affinities, and verified that everything is working great. Thanks for the improvement!

HN23 (approved):

PR #362, opened 2024-04-11
mv license into global parameters

darrylweaver:
Move the license secret into a global parameter so that the feeds subchart can also use this value.
Also took the opportunity to move the imagePullSecret.

### DIFF
--- a/stable/enterprise/values.yaml
+++ b/stable/enterprise/values.yaml
@@ -11,6 +11,18 @@ global:
   ##
   nameOverride: ""
 
+  ## @param licenseSecretName Name of the Kubernetes secret containing your license.yaml file
+  ## This must be manually created. For example with the following command:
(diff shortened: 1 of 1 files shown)


Btodhunter:
Thanks for the PR! This is definitely a good call out, however I don't think it's addressed properly in this PR. The templates that use the imagePullSecret & imagePullPolicy need to be updated to utilize `global.imagePullSecret` & `global.imagePullPolicy`. The feeds chart should also be updated to match these conventions.

We also have a policy that breaking changes to the values file require a major version bump on the chart. We try to batch our breaking changes so we can reduce the impact to customers of a major version bump. On that note, we're currently working on a major version update for the chart, so we will include these changes into that release.

Closing this PR as it will be addressed in our next major version release of the chart.

PR #363, opened 2024-04-11
adding missingRegistryOverride config for k8s inventory

HN23:


stable/k8s-inventory/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: k8s-inventory
-version: 0.2.2
+version: 0.2.3
 appVersion: "1.2.0"
```
bradleyjones, line 4: we need to bump this to 1.4.0 and bump the image used as well I think

bradleyjones (changes requested):

bradleyjones (approved):

PR #373, opened 2024-04-24
fix: update with new available log levels

juanjsebgarcia:


HN23:
not needed as its in the 5.5.0 branch

PR #376, opened 2024-05-10
fix: allow override of entire db_connect_args, not just single values

zhill:


stable/enterprise/files/default_config.yaml
```diff
@@ -52,13 +52,19 @@ credentials:
     host: "${ANCHORE_DB_HOST}"
     port: "${ANCHORE_DB_PORT}"
     name: "${ANCHORE_DB_NAME}"
+    {{- if .Values.anchoreConfig.database.dbConnectArgs }}
```
Btodhunter, line 55: To match the conventions of the rest of the chart, this `if` & corresponding `end` statement should only be indented 2 spaces.

Btodhunter: We should also add `anchoreConfig.database.dbConnectArgs` to our values file with a corresponding @param decoration (and update the README)

Btodhunter (approved):

PR #377, opened 2024-05-17
bumping enterprise image to 5.5.1

HN23:

PR #379, opened 2024-05-20
updating enterprise feeds dep lock

HN23:


HN23:
updated in https://github.com/anchore/anchore-charts/pull/377

Issue #221, opened 2022-10-17
connectionPoolSize/connectionPoolMaxOverflow for engine-upgrade job


zlugo:
Hi, I am trying to use Anchore with a PostgreSQL database that does not support lots of open connections.
Therefore I've decreased the `dbConfig.connectionPoolSize` and `dbConfig.connectionPoolMaxOverflow` values which works very well in general and avoids exhausting the available database connections.

However, the pods created for the engine-upgrade job don't take up these values. Instead the logs show the default values for both the pool size and the max overflow, e.g.:
`[MainThread] [anchore_manager.util.db/connect_database()] [INFO] DB params: {"db_connect_args": {"connect_timeout": 86400}, "db_pool_size": 30, "db_pool_max_overflow": 100}`

I assume this is the reason the engine-upgrade pods are crashing with
`[MainThread] [anchore_manager.util.logging/log_error()] [ERROR] Error: DB connection failed - exception: %stest connection failed - exception: (psycopg2.OperationalError) FATAL:  remaining connection slots are reserved for non-replication superuser connections`

Is there a way to make this work, i.e. by mapping the config-volume the same way as it is done in e.g. the api_deployment.yaml?

PR #385, opened 2024-05-31
bumping feeds chart image (#221)

HN23:
* bumping feeds chart image
* updating chart postgres dep to latest minor version, update readme and tests

---------


Btodhunter (approved):
lgtm

PR #395, opened 2024-08-09
enterprise: bumping image version to v5.8.1

HN23:
feeds will need to be merged first, then update the dependency/lock on enterprise before this is merged.
https://github.com/anchore/anchore-charts/pull/394


zhill (approved):
lgtm!

PR #404, opened 2024-09-13
ENTERPRISE-4495: Increase default polling interval

gnyahay:


Btodhunter (approved):
lgtm

HN23:
is there a reason we cant just use a values file to bump the interval?

PR #410, opened 2024-10-08
bug: fix default vuln values

### LINKED ISSUE #409: Anchore Engine Chart - invalid default values
In the v5.10.0 release a section was added to the default helm values that is invalid

```
anchoreConfig:
  policy_engine:
# vvv this part was added vvv
    vulnerabilities:
      matching:
        exclude:
          providers: null
          package_types: null
```

This conflicts with the requirements of the anchore_configmap template where those cant be null.

```
{{- $exclude_providers := required "anchoreConfig.policy_engine.vulnerabilities.matching.exclude.providers is required" .Values.anchoreConfig.policy_engine.vulnerabilities.matching.exclude.providers -}}
{{- $exclude_package := required "anchoreConfig.policy_engine.vulnerabilities.matching.exclude.package_types is required" .Values.anchoreConfig.policy_engine.vulnerabilities.matching.exclude.package_types -}}
```

The code expects a list there so the null values can be changed to empty lists `[]` and the base values will deploy. 

new values:

```
anchoreConfig:
  policy_engine:
    vulnerabilities:
      matching:
        exclude:
          providers: []
          package_types: []
```

will be submitting PR to resolve.

misterjacko:
Resolves #409

### DIFF
--- a/stable/enterprise/values.yaml
+++ b/stable/enterprise/values.yaml
@@ -600,8 +600,8 @@ anchoreConfig:
     vulnerabilities:
       matching:
         exclude:
-          providers: null
-          package_types: null
+          providers: []
+          package_types: []
 
     ## @param anchoreConfig.policy_engine.enable_user_base_image Enables usage of Well Known Annotation to identify base image for use in ancestry calculations
     enable_user_base_image: true


Btodhunter:
@bdelvecchio @misterjacko Thanks for the PR! However, we have these values default to `null` on purpose so that we can check existing deployments for exclusions and fail with an error for existing users to update this value. Our template will change `null` to `[]` if the value is not changed or if it is a new deployment. We will default this to `[]` when we do the next major version bump of the chart.

PR #411, opened 2024-10-09
use empty array for these values

bdelvecchio:
I got the following error when applying the helm chart:

```
│ Error: execution error at (enterprise/templates/simplequeue_deployment.yaml:21:39): anchoreConfig.policy_engine.vulnerabilities.matching.exclude.providers is required
```

Changing the default value from `null` to `[]` (Empty array) fixes the issue.


bdelvecchio:
closing as duplicate of #410

PR #414, opened 2024-10-24
DEVOPS-640: Check that PGDATA and postgres mountPoint are in alignmen…

gnyahay:
Check that PGDATA and postgres mountPoint are in alignment if configured

### DIFF
--- a/stable/enterprise/Chart.yaml
+++ b/stable/enterprise/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: enterprise
-version: "3.0.2"
+version: "3.0.3"
 appVersion: "5.10.0"
 kubeVersion: 1.23.x - 1.30.x || 1.23.x-x - 1.30.x-x
 description: |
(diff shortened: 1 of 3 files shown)


Btodhunter:
@gnyahay Thanks for the PR! This is a great thing to validate. I wonder if this would be better suited for the upstream Bitnami postgresql chart though? It would fit in pretty well right here - https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/_helpers.tpl#L320

gnyahay:
> @gnyahay Thanks for the PR! This is a great thing to validate. I wonder if this would be better suited for the upstream Bitnami postgresql chart though? It would fit in pretty well right here - https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/_helpers.tpl#L320

I thought about that. They may have their reasons for keeping these values separate. If the variable for the data directory was based off the variable for the mount point in the upstream chart then this wouldn't even be needed. They may have their reasons for setting the values independently and so my thought was that this validation is specific to the use case of their chart in our chart and therefore it makes sense to go into our chart.

If I put it in their chart I would probably need to include the same logic for replicas as well, not just primary.

PR #417, opened 2024-11-06
feat: make log_level and malware api configurable

juanjsebgarcia:
Add the initial API configurable configs of log_level and malware


HN23 (approved):
lgtm

PR #418, opened 2024-11-14
Add image registry/repository for bitnami charts and storageclass

gnyahay:
We often get asked how to point to images in a different registry and its very unclear from the values.yaml how one would go about doing that. The goal of this PR is to make it more obvious.


Btodhunter (approved):
lgtm

PR #423, opened 2024-11-27
Update values.yaml for resource configuration adjustments

Sopuru:
Updated values here to match with our recommendations: [https://github.com/anchore/support-tools/blob/main/customer-onboarding/enterprise-chart-values.yaml](https://github.com/anchore/support-tools/blob/main/customer-onboarding/enterprise-chart-values.yaml)
[](url)

### DIFF
--- a/stable/enterprise/Chart.yaml
+++ b/stable/enterprise/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: enterprise
-version: "3.2.0"
+version: "3.2.1"
 appVersion: "5.12.0"
 kubeVersion: 1.23.x - 1.31.x || 1.23.x-x - 1.31.x-x
 description: |
--- a/stable/enterprise/values.yaml
+++ b/stable/enterprise/values.yaml
@@ -801,7 +801,7 @@ analyzer:
   ##
   resources: {}
     # requests:
-    #   cpu: 2
+    #   cpu: 1
     #   memory: 8G
     # limits:
     #   memory: 8G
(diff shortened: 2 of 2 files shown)


lwade:
This brings the resourcing into line with our recommended requirements

lwade:
@Sopuru  can you cross-reference with the docs please: https://docs.anchore.com/current/docs/deployment/requirements/ ? Specifically:

> Requests specify the desired resource amounts for the container, while limits specify the maximum resource amounts the container is allowed. We have found that setting the request and limit to the same value provides the best quality of service (QoS) from Kubernetes.

We do not recommend setting less than 1 CPU unit for any containers. Less than this could result in unexpected behaviour and should only be used in testing scenarios.

For the catalog, policeEngine and postgresql containers, we recommend a minimum of 2 CPU units.

We do not recommend setting memory units to less than 8G except for API and UI services, where we recommend starting at 4G. Less than these values could result in OOM errors or containers restarting unexpectedly.

Sopuru:
@lwade Corrections have been made. Checks are on going. Please review again after the checks

lwade:
Almost there I think: 

1. Analyzer should be `cpu: 1`
2. osaaMigrationJob doesn't need adjusting I don't think (not a core service)

Sopuru:
Yeah! Corrected now

lwade:
LGTM!

lwade:
@Sopuru can you rebase and merge this please? Thanks! :)

Btodhunter (approved):
lgtm

PR #425, opened 2024-12-12
anchore-enterprise 5.13 add max_scan_time for analyzer malware

HN23:


stable/enterprise/tests/__snapshot__/configmap_test.yaml.snap
```diff
@@ -8,6 +8,7 @@ should render the configmaps:
           clamav:
             db_update_enabled: true
             enabled: <ALLOW_API_CONFIGURATION>
+            max_scan_time: 600000
```
bobmel, line 11: With the latest code change I'm making the new default will be 180000 
 (= 30 minutes) rather than 600000 (= 10 minutes).
 
 Probably the helm chart should have that as default too?

stable/enterprise/values.yaml
```diff
@@ -503,6 +503,8 @@ anchoreConfig:
         clamav:
           enabled: "<ALLOW_API_CONFIGURATION>"  # false
           db_update_enabled: true
+          # Maximum time in milliseconds that a ClamAV scan is allowed to run
+          max_scan_time: 600000
```
bobmel, line 507: Dito.

bobmel (approved):

PR #430, opened 2025-01-13
Increase resource limit comments/hints

gnyahay:


lwade (dismissed):
LGTM

DreadPirateDuo (approved):
Looks good

stable/enterprise/values.yaml
```diff
@@ -814,12 +815,14 @@ analyzer:
...
   ## Commented values below are just a suggested baseline. Contact Anchore support for deployment specific recommendations.
+  ## Please note that we do not suggest setting CPU limits.
   ##
   resources: {}
     # requests:
-    #   cpu: 1
+    #   cpu: 2
     #   memory: 8G
     # limits:
+    #   cpu: 2
```
zhill, line 825: It seems that if we don't suggest setting CPU limits, then we shouldn't have them in here at all, just the requests.

stable/enterprise/values.yaml
```diff
@@ -1012,6 +1019,7 @@ catalog:
 #########################################################
 dataSyncer:
   ## @param dataSyncer.replicaCount Number of replicas for the Anchore DataSyncer deployment
+  ## Please note that we do not suggest more than 1 instance of dataSyncer.
```
zhill, line 1022: If we don't suggest this, should we remove this as a configurable option? I think we have to assume that while our comments should be correct, they won't be read nor followed.

stable/enterprise/values.yaml
```diff
@@ -1442,13 +1460,15 @@ simpleQueue:
...
   ##
   resources: {}
     # requests:
-    #   cpu: 1
+    #   cpu: 2
     #   memory: 8G
     # limits:
-    #   memory: 8G
+    #   cpu: 4
+    #   memory: 12G
```
zhill, line 1471: Is this really intended for the queue? 12GB seems high

zhill (changes requested):
I appreciate the work here and like the intent of this, but I think our general approach for documenting best practices etc needs discussion around what is the right level of info in the values.yaml itself compared to other documentation that covers how to use the chart to achieve specific goals.

I agree the comments should be correct and any guidance we give shouldn't be inconsistent. I think there is more discussion on the specific values chosen, and the intent of those specific settings so we can review in that context.

stable/enterprise/values.yaml
gnyahay: In cases where they are going to set them anyway I want to make sure they don't set it too low.
I am also hoping the BB chart follows our example: https://repo1.dso.mil/big-bang/product/packages/anchore-enterprise/-/blob/main/chart/values.yaml?ref_type=heads#L1444

stable/enterprise/values.yaml
gnyahay: I could lower it if you prefer...

stable/enterprise/values.yaml
gnyahay: I could lower it if you prefer.

gnyahay: I have reverted this back to 8GB

stable/enterprise/values.yaml
lwade: Yes please! @gnyahay @InfraInnovator @zhill - lets just remove the option

gnyahay: You can actually run more than one dataSyncer, correct?
There are actually REST endpoints made by other services (policy I think) to dataSyncer, correct?
One might want to have more than 1 for that reason...

PR #432, opened 2025-01-16
Allow configuring ui custom_message via chart

gnyahay:


stable/enterprise/values.yaml
```diff
@@ -745,6 +745,13 @@ anchoreConfig:
       # admin: True
       # standard: True
 
+    ## @param anchoreConfig.ui.custom_message.title A title key, whose string value provides a title for the message
+    ## @param anchoreConfig.ui.custom_message.message A message key, whose string value is the message itself
+    ##
+    custom_message: {}
```
holt, line 751: Probably best to comment this out because I think our validator will expect the keys to be provided if it's present.

gnyahay: Mike, that's a good shout! I just tested it and if custom_message is empty nothing gets rendered in the config map which represents the config-ui.yaml.

HN23 (approved):

Btodhunter (approved):

PR #446, opened 2025-03-05
Removing policy engine feed sync cycle timers (now on data syncer)

lwade:
policy engine feed sync cycle timers are no longer used.


lwade:
Looks like removing stuff needs a bump in the chart version?

lwade:
see discussion here: https://anchore.atlassian.net/wiki/spaces/ENG/pages/716242945/Hosted+Feeds+Anchore+Data+Service

stable/enterprise/values.yaml
```diff
@@ -600,14 +600,6 @@ anchoreConfig:
     ##
     down_analyzer_task_requeue: true
 
-  policy_engine:
-    ## @param anchoreConfig.policy_engine.cycle_timers.feed_sync Interval to run a feed sync to get latest cve data
-    ## @param anchoreConfig.policy_engine.cycle_timers.feed_sync_checker Interval between checks to see if there needs to be a task queued
-    ##
-    cycle_timers:
-      feed_sync: 14400
-      feed_sync_checker: 3600
```
asomya, line 609: References to `policy_engine.cycle_timers` also needs to be removed from all files. It's present in default_config.yaml, ossa_config.yaml, mappings etc.

stable/enterprise/values.yaml
```diff
@@ -600,14 +600,6 @@ anchoreConfig:
     ##
     down_analyzer_task_requeue: true
 
-  policy_engine:
```
bot190, line 603: This line needs to stick around

lwade: whoops, good catch, didn't even see I'd deleted it!

Btodhunter:
LGTM! Thanks for the update @lwade 

I reverted all the changes to mappings.py because that file is specifically for the values conversion script that is only used for migrating from the anchore-engine chart to the enterprise chart. It needs to have the same values as the Enterprise values file referenced in the migration docs.

I also updated the config snapshots for the Helm unit tests to get them passing.

Btodhunter (approved):

lwade:
Thank you @Btodhunter !

PR #447, opened 2025-03-20
Added imageSpec for registry/repo:tag overrides

blang9238:

### DIFF
--- a/stable/enterprise/templates/analyzer_deployment.yaml
+++ b/stable/enterprise/templates/analyzer_deployment.yaml
@@ -39,7 +39,10 @@ spec:
         {{- include "enterprise.common.cloudsqlContainer" . | nindent 8 }}
       {{- end }}
         - name: "{{ .Chart.Name }}-{{ $component | lower }}"
-          image: {{ .Values.image }}
+          image: {{- if and .Values.imageSpec.registry .Values.imageSpec.repository .Values.imageSpec.tag }} {{ printf "%s/%s:%s" .Values.imageSpec.registry .Values.imageSpec.repository .Values.imageSpec.tag }}
+                 {{- else }}
+                   {{ .Values.image }}
+                 {{- end }}
           imagePullPolicy: {{ .Values.imagePullPolicy }}
         {{- with .Values.containerSecurityContext }}
           securityContext:
--- a/stable/enterprise/values.yaml
+++ b/stable/enterprise/values.yaml
@@ -21,6 +21,14 @@ global:
 ##
 image: docker.io/anchore/enterprise:v5.15.0
 
+## @oaram imageSpec ImageSpec used for all Anchore Enterprise deployments by separating out registry/repo:tag, excluding Anchore UI 
+## (only one of image or imageSpec should be set)
+##
+imageSpec: {}
+  # registry: docker.io
+  # repository: anchore/enterprise
+  # tag: v5.15.0
+
 ## @param imagePullPolicy Image pull policy used by all deployments
 ## ref: https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy
 ##
@@ -1484,6 +1492,14 @@ ui:
   ## @param ui.image Image used for the Anchore UI container
   ##
   image: docker.io/anchore/enterprise-ui:v5.15.0
+
+  ## @param ui.imageSpec imageSpec used for the Anchore UI container
+  ## (only one of image or imageSpec should be set)
+  ##
+  imageSpec: {}
+    # registry: docker.io
+    # repository: anchore/enterprise-ui
+    # tag: v5.15.0
 
   ## @param ui.imagePullPolicy Image pull policy for Anchore UI image
   ##
--- a/stable/enterprise/templates/api_deployment.yaml
+++ b/stable/enterprise/templates/api_deployment.yaml
@@ -37,7 +37,10 @@ spec:
         {{- include "enterprise.common.cloudsqlContainer" . | nindent 8 }}
       {{- end }}
         - name: "{{ .Chart.Name }}-{{ $component | lower }}"
-          image: {{ .Values.image }}
+          image: {{- if and .Values.imageSpec.registry .Values.imageSpec.repository .Values.imageSpec.tag }} {{ printf "%s/%s:%s" .Values.imageSpec.registry .Values.imageSpec.repository .Values.imageSpec.tag }}
+                 {{- else }}
+                  {{ .Values.image }}
+                {{- end }}
           imagePullPolicy: {{ .Values.imagePullPolicy }}
         {{- with .Values.containerSecurityContext }}
           securityContext:
--- a/stable/enterprise/templates/catalog_deployment.yaml
+++ b/stable/enterprise/templates/catalog_deployment.yaml
@@ -43,7 +43,10 @@ spec:
         {{- include "enterprise.common.cloudsqlContainer" . | nindent 8 }}
       {{- end }}
         - name: "{{ .Chart.Name }}-{{ $component | lower }}"
-          image: {{ .Values.image }}
+          image: {{- if and .Values.imageSpec.registry .Values.imageSpec.repository .Values.imageSpec.tag }}  {{ printf "%s/%s:%s" .Values.imageSpec.registry .Values.imageSpec.repository .Values.imageSpec.tag }}
+                 {{- else }}
+                   {{ .Values.image }}
+                 {{- end }}
           imagePullPolicy: {{ .Values.imagePullPolicy }}
         {{- with .Values.containerSecurityContext }}
           securityContext:
--- a/stable/enterprise/templates/datasyncer_deployment.yaml
+++ b/stable/enterprise/templates/datasyncer_deployment.yaml
@@ -35,7 +35,10 @@ spec:
         {{- include "enterprise.common.cloudsqlContainer" . | nindent 8 }}
       {{- end }}
         - name: "{{ .Chart.Name }}-{{ $component | lower }}"
-          image: {{ .Values.image }}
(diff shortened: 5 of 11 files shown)


zhill:
@blang9238 what is the reason for this PR?

gnyahay (approved):
It appears to be backwards compatible.

blang9238:
@zhill P1 asked if we could support separating out the registry / repo / tag (similar to other charts) but mentioned that BB will only pull it down if it's in our upstream. I'm reworking this currently to use the common.tpl. I'll send you the internal discussion.

stable/enterprise/templates/analyzer_deployment.yaml
```diff
@@ -39,7 +39,11 @@ spec:
         {{- include "enterprise.common.cloudsqlContainer" . | nindent 8 }}
       {{- end }}
         - name: "{{ .Chart.Name }}-{{ $component | lower }}"
-          image: {{ .Values.image }}
+          image: {{- if and .Values.imageSpec.registry .Values.imageSpec.repository .Values.imageSpec.tag -}}
+                    {{ printf "%s/%s:%s" .Values.imageSpec.registry .Values.imageSpec.repository .Values.imageSpec.tag | trim | indent 1}}
```
zhill, line 43: this doesn't, for example handle digests instead of a tag.

zhill: but similar bitnami format does support digest as well in the object via the "Digest" field, so not sure if that needs to also be supported here as well.

blang9238: Good point, can certainly add it.

stable/enterprise/templates/_common.tpl
```diff
@@ -247,6 +247,53 @@ app.kubernetes.io/component: {{ $component | lower }}
 {{- end -}}
 
 
+{{/*
+Create an image specification template that can override the default image
+based on global settings
+*/}}
+{{- define "enterprise.common.imageOverride" -}}
```
zhill, line 254: "Override" is confusing here, it should just be "enterprise.common.image".

stable/enterprise/templates/_common.tpl
```diff
@@ -247,6 +247,53 @@ app.kubernetes.io/component: {{ $component | lower }}
...
+{{- else -}}
+  {{ .Values.image | quote }}
+{{- end }}
+{{- end }}
+
+{{/*
+Create an image specification template for the UI that can override the default image
+based on component-specific or global settings
+*/}}
+{{- define "enterprise.ui.imageOverride" -}}
```
zhill, line 269: Same here as above, just use "enterprise.ui.image".

stable/enterprise/templates/_common.tpl
```diff
@@ -247,6 +247,53 @@ app.kubernetes.io/component: {{ $component | lower }}
...
+  {{ .Values.ui.image | quote }}
+{{- end }}
+{{- end }}
+
+
+{{/*
+Create an image specification template for the UI that can override the default image
+based on component-specific or global settings
+*/}}
+{{- define "enterprise.kubectl.imageOverride" -}}
```
zhill, line 285: And here.

stable/enterprise/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: enterprise
-version: "3.5.2"
+version: "3.5.3"
```
zhill, line 3: Since this is a new feature, though backwards compatible, I think this is "3.6.0" not a patch update.

zhill (changes requested):
@blang9238 I also checked on how we can make this a bit easier for the user and something like this snippet would allow the 'image' value to be either a map or a string so we wouldn't need different keys:
```
{{- define "enterprise.common.image" -}}
# Check if the 'image' is a string or not, if not, use it as a map.
{{- if eq (printf "%T" .Values.image) "string" }}
  {{- .Values.image | trim -}}
{{- else -}}
  # Assume a map (but verify non-nil..) and construct string from the parts.
  {{- printf "%s/%s:%s" .Values.image.Registry .Values.image.Repository Values.image.Tag | trim -}}
{{- end -}}
{{- end }}
```

I'd like others' opinions on if allowing different types there would be more confusing, but at least it is a single place in the Values to set the images rather than 2 that interact. @HN23 or @Btodhunter, thoughts?

stable/enterprise/templates/_common.tpl
blang9238: Ack, so should it be:

image: docker.io/anchore/enterprise:v5.16.0
  # registry: docker.io
  # repository: anchore/enterprise
  # tag: v5.16.0
  # digest: sha256:12345abcde
  
Then if registry / repo / tag / digest are defined, use that, otherwise default to just the pullstring?

or should I split it out into something like:

image: 
  pullstring: docker.io/anchore/enterprise:v5.16.0
  # registry: docker.io
  # repository: anchore/enterprise
  # tag: v5.16.0
  # digest: sha256:12345abcde
  
And then do the same checks but modify .Values.image to .Values.image.pullstring?

blang9238: Disregard, saw your later comment about checking the 'image' value either being a map or a string

stable/enterprise/Chart.yaml
blang9238: @zhill since v5.16.0 took the chart to 3.6.0, should I then bump this to 3.7.0?

blang9238:
@zhill pushed some updates for your review.

stable/enterprise/templates/_common.tpl
```diff
@@ -247,6 +247,50 @@ app.kubernetes.io/component: {{ $component | lower }}
...
 
 
+{{/*
+Create an image specification template that can override the default image
+based on global settings
+*/}}
+{{- define "enterprise.common.image" -}}
+{{- if eq (printf "%T" .Values.image) "string" }}
+  {{- .Values.image | trim -}}
+{{- else if and .Values.image.digest -}}
```
zhill, line 257: Is the "and" necessary here? I may be wrong but I don't think the "and" is helpful here.

stable/enterprise/templates/_common.tpl
```diff
@@ -247,6 +247,50 @@ app.kubernetes.io/component: {{ $component | lower }}
...
+{{- end }}
+
+{{/*
+Create an image specification template for the UI that can override the default image
+based on component-specific or global settings
+*/}}
+{{- define "enterprise.ui.image" -}}
+{{- if eq (printf "%T" .Values.ui.image) "string" }}
+  {{- .Values.ui.image | trim -}}
+{{- else if and .Values.ui.image.digest -}}
```
zhill, line 271: same "and" question

stable/enterprise/templates/_common.tpl
```diff
@@ -247,6 +247,50 @@ app.kubernetes.io/component: {{ $component | lower }}
...
 
+{{/*
+Create an image specification template that can override the default image
+based on global settings
+*/}}
+{{- define "enterprise.common.image" -}}
+{{- if eq (printf "%T" .Values.image) "string" }}
+  {{- .Values.image | trim -}}
+{{- else if and .Values.image.digest -}}
+  {{- printf "%s/%s@%s" .Values.image.registry .Values.image.repository .Values.image.digest | trim -}}
```
zhill, line 258: in both these cases, tag or digest, if the user only fills out part of the dict, it will output a bad reference that will fail to run the pods. I'd like to see some checks that the registry and repository properties exist and are non-nil.

zhill (changes requested):
Looks like there is a breaking change, so we need to fix that (the new way for kubectl works but is breaking if a user had set it previously). Otherwise just a few questions and minor value checks to ensure the image refs are constructed well. Overall looking good. Getting close!

stable/enterprise/templates/_common.tpl
blang9238: Ack thanks @zhill, I'll revisit and get checks in place for this.

stable/enterprise/templates/_common.tpl
blang9238: nope, removed.

stable/enterprise/templates/_common.tpl
blang9238: Ok this should handle it better.

stable/enterprise/templates/_common.tpl
blang9238: removed.

blang9238:
@zhill Is this aligned with what you were thinking? I left the trims in, renders correctly when I tested locally and with helm templating.

HN23:
@blang9238 
Can you update the stable/enterprise/templates/tests/anchorectl_smoketest.yaml as well. its currently {{ .Values.image }}, which causes it to not be right if passing in a dict.

stable/enterprise/values.yaml
```diff
@@ -17,9 +17,16 @@ global:
 ## Common params used by all Anchore k8s resources
 ###################################################
 
-## @param image Image used for all Anchore Enterprise deployments, excluding Anchore UI
+## @param imageOverride imageOverride used for all Anchore Enterprise deployments by separating out registry/repo:tag, excluding Anchore UI
```
HN23, line 20: is imageOverride an old/deleted value? If so, we should remove mentions of it.

stable/enterprise/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: enterprise
-version: "3.8.0"
+version: "3.8.1"
```
HN23, line 3: We should probably bump this to the next minor version instead of a patch since its changing/adding a new way the chart works. Also we should probably update the README to include the changes (both the kubectlImage change and the image/ui.image change)

HN23 (approved):
lgtm. Thanks for the PR @blang9238  !

zhill (approved):
lgtm

PR #450, opened 2025-04-04
Make cloudsql proxy a sidecar during migration

gnyahay:


gnyahay:
UPDATE: Should be fixed but I should probably add a test case for that

gnyahay:
Using https://github.com/anchore/anchore-charts/pull/487 where I have cleaned up the commits a bit

PR #459, opened 2025-05-09
Update envvars_configmap.yaml

lwade:
change default for new installs - must then be explicitly enabled in order to use  "packages" gate and "verify" trigger


HN23:
Hey @lwade, thanks for the PR. This change will actually impact existing deployments as well as you're changing a default value. Unfortunately, we wont be able to make this change without further discussions and possibly a major version bump.

lwade:
@HN23 want me to close this one in respect of #464 ?

HN23:
> @HN23 want me to close this one in respect of #464 ?

Yes please. Thanks @lwade

PR #461, opened 2025-05-14
Enterprise dependent chart pullcreds

HN23:
This helps get around dockerhub's rate limiting. If you have credentials to pull enterprise, we can use those same credentials to pull the redis and postgresql images. 

If the customer is already overriding it due to running into the rate limiting issues, then their override would already take priority over whats in the default values.yaml.


InfraInnovator (dismissed):
LGTM

InfraInnovator (approved):
Looks good

PR #465, opened 2025-05-21
add serviceAccount

luismacosta:
add serviceAccount


HN23:
Hi @luismacosta , Thanks for the contribution! I left a couple of comments. We'll also want to bump the chart version so linting and such will pass.

luismacosta:
> Hi @luismacosta , Thanks for the contribution! I left a couple of comments. We'll also want to bump the chart version so linting and such will pass.

chart version updated

PR #474, opened 2025-06-12
Render endpoint hostnames as a formal FQDN

gnyahay:
Our hostnames in a kubernetes environment should be treated as an FQDN but because they do not end in a dot they are treated as partial hostnames. When a DNS lookup is performed for service to service communication 5 lookups are required. By the 5th lookup it is assumed to be an FQDN. By adding a trailing dot only 1 lookup will be required. There is still an ability to overwrite the domainSuffix in the event someone needs the former behavior.

### DIFF
--- a/stable/enterprise/Chart.yaml
+++ b/stable/enterprise/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: enterprise
-version: "3.10.0"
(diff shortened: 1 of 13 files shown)


gnyahay:
I broke this one... See https://github.com/anchore/anchore-charts/pull/478

PR #475, opened 2025-06-13
enterprise: Fix configmap unittests

HN23:
These two tiny changes fixes the snapshot unittests for configmaps so that they're actually useful/readable.


InfraInnovator (approved):
LGTM

PR #478, opened 2025-06-23
Render endpoint hostnames as a formal FQDN

gnyahay:


gnyahay:
Replacement for https://github.com/anchore/anchore-charts/pull/474

gnyahay:
Tested NO_PROXY/no_proxy
[12:05](https://anchore.slack.com/archives/D07GUU5SWDB/p1750694758102869)
so far it seems the following is needed in no_proxy/NO_PROXY:
anchore.svc.cluster.local.
(i.e. including the trailing dot)
Presumably it could be included with and without it
12:06
something of note is that i have minio configured for object store and i needed it without the trailing dot:
minio.anchore.svc.cluster.local

gnyahay:
> Thanks for the PR @gnyahay! Appreciate any improvements to the chart. This one in particular is tricky because it requires manual intervention for customers who utilize NO_PROXY. Based on the guidance in our readme, this would require a major version bump and could cause potential downtime for customers if they upgraded without reading the release notes. https://github.com/anchore/anchore-charts/blob/main/stable/enterprise/README.md#release-notes
> 
> > Major Chart Version Change (e.g., v0.1.2 -> v1.0.0): Signifies an incompatible breaking change that necessitates manual intervention, such as updates to your values file or data migrations.
> 
> I did a little bit of experimentation and found that adding the following to `enterprise.common.podSpec` significantly reduces DNS queries:
> 
> ```
> dnsConfig:
>   options:
>     - name: ndots
>       values: '2'
> ```
> 
> I _think_ this would fix the problem without requiring any manual intervention for our customers. It should also reduce the number of DNS requests made by the UI and analyzer services when reaching out to registries/db/redis. I want to think about other potential blowback from this change before implementing it though. The things that come top of mind are external DB/Redis hostnames as well as internal registry hostnames. I'd imagine most folks would use FDQN, `servicename`, or `servicename.namespace` which should still initiate lookups on the search domains.

I'm fine with that

Btodhunter:
Closing this PR - a solution was implemented in #486 that allows lowers the default pod DNS ndot config to reduce unnecessary  lookups.

PR #483, opened 2025-06-30
enterprise: change reports_worker default value

HN23:


svietry (approved):
This looks correct.  Thanks

PR #501, opened 2025-08-07
Added templating for audit resource uris to allow specifying additional uris

blang9238:

### DIFF
--- a/stable/enterprise/values.yaml
+++ b/stable/enterprise/values.yaml
@@ -320,9 +320,12 @@ anchoreConfig:
     timeout_keep_alive: 5
 
   ## @param anchoreConfig.audit.enabled Enable audit logging
+  ## @param anchoreConfig.audit.additionalResourceURIs Additional resource URIs to audit
   ##
   audit:
     enabled: true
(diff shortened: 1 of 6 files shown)


stable/enterprise/values.yaml
```diff
@@ -320,9 +320,12 @@ anchoreConfig:
     timeout_keep_alive: 5
 
   ## @param anchoreConfig.audit.enabled Enable audit logging
+  ## @param anchoreConfig.audit.additionalResourceURIs Additional resource URIs to audit
   ##
   audit:
     enabled: true
+    additionalResourceURIs: {}
```
HN23, line 327: Should this be a list?

HN23 (commented):
We'll need to bump the chart.yaml too.

stable/enterprise/values.yaml
blang9238: oh yep, that's a good point.

blang9238:
Yep sorry @HN23, bumped the chart version and changed placeholder to empty array.

Issue #511, opened 2025-09-03
Need ability to create sidecar containers to anchore deployments.


gaevans-ebay:
We have a use-case that requires the addition of a forward proxy to route traffic from the anchore analyzers and catalogs to our internal image registries we could set up another deployment and route trough that but it is an non ideal solution as it creates a second hop outside of the security zone of the pod. a better solution is to run a sidecar proxy that we route traffic to and then use that to determine how to modify the requests through whichever of the proxies are needed for the appropriate endpoints.

PR #531, opened 2025-11-12
Adds support for sidecars to address #511

blang9238:


zhill (changes requested):
Overall it looks good, but I'd like to change the name of the property in the values.yaml from 'sidecars' to 'initContainers' so that it maps directly to the actual k8s concept and also makes it clear that's how users can add non-sidecar init containers.

HN23 (approved):

zhill (approved):
thanks for the naming change! lgtm

PR #532, opened 2025-11-12
encode appdb uri

HN23:
Encodes the secrets used by ui for anchore_appdb_uri and anchore_redis_uri. This only encodes it for instances where the helm chart is managing the secret. If you're using existing secrets, you will need to do the encoding by yourself.


stable/enterprise/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: enterprise
-version: "3.17.2"
+version: "3.17.3"
```
HN23: not released yet. this is only going into our environment, so we dont bump it to 3.18.0 yet.

PR #593, opened 2026-04-11
Moved AWS S3 credentials to a secret, adds existing secret support (#978)

blang9238:


stable/enterprise/templates/storage_credential_secret.yaml
```diff
@@ -0,0 +1,39 @@
+{{/*
```
hungnguyen-anchore, line 1: Hey @blang9238 thanks for the work on this. I think this current implementation presents a bug. Im thinking through the case of when a user wants to do an osaaMigration, so they enable the osaaMigrationJob and put their keys in the objectStoreMigration.storage_driver.config for the migration. I dont think the migration credentials ever make it to the migration pod with this approach.

hungnguyen-anchore (approved):
lgtm! thanks!

stable/enterprise/templates/storage_credential_secret.yaml
blang9238: Thanks Hung, from the Slack discussion, this appears to be fine.

PR #613, opened 2026-06-29
adding support for encryption feature

hungnguyen-anchore:


svietry (approved):

stable/enterprise/values.yaml
```diff
@@ -447,6 +447,12 @@ anchoreConfig:
   ## @param anchoreConfig.database.dbConnectArgs Set custom database connection args (legacy/psycopg2); If specified, this overrides other database connection settings
   ## @param anchoreConfig.database.ngEngineArgs Set custom database engine arguments (ng/psycopg3); If specified, this overrides pool_size and max_overflow for ng services
   ## @param anchoreConfig.database.ngDbConnectArgs Set custom database connection args (ng/psycopg3); If specified, this overrides other database connection settings for ng services
+  ## @param anchoreConfig.database.disable_db_encryption_unsafe Disable database encryption at rest. UNSAFE - intended for testing/migration only.
```
juanjsebgarcia: I'm not sure we want to limit them to `testing/migration only` they can run with it on if they want but we should say its not recommended.

juanjsebgarcia:
```suggestion
  ## @param anchoreConfig.database.disable_db_encryption_unsafe Disable encryption of secrets within the database. UNSAFE.
```

juanjsebgarcia: And whilst it is encryption at rest, it's also encryption during a running session. As in the values are encrypted Strings which then may also be on an encrypted volume. Does that make sense?

PR #629, opened 2026-08-06
feat: expose UI `log_format` setting in the enterprise chart

holt:
## Summary

Adds `anchoreConfig.ui.log_format` to the enterprise chart (`text` default, `json` for log aggregation pipelines), rendered into `config-ui.yaml` alongside the existing `log_level`. This exposes the structured logging capability arriving with the enterprise-ui 6.2 release line (anchore/anchore-on-prem-ui ENTERPRISE-8841, epic ENTERPRISE-8852) through the Helm chart, per the original RFE (ENTERPRISE-5947) request for chart-level configuration.

- Mirrors the `log_level` plumbing exactly: `@param` annotation in `values.yaml`, `squote`d render in `templates/ui_configmap.yaml`, README parameter row
- Chart version bumped `3.25.2` -> `3.26.0` (minor, per new-parameter precedent)
- New unit test asserts the value renders into `config-ui.yaml`; configmap snapshot regenerated (diff is the single expected line)

## Safety with pre-6.2 UI images

The key is safe to render unconditionally against every enterprise-ui image generation: the modern config loader iterates its own metadata registry and reads known settings from the parsed YAML (unknown keys are never enumerated), and the older-generation loader likewise reads specific properties from the `yaml.safeLoad` result. In all cases an unrecognized `log_format` key is inert. With a 6.2+ image, the `text` default is byte-identical to current behavior, and an invalid value falls back to `text` with a validation warning rather than failing startup.

## Testing

`helm unittest` (helm v3.8.0, helm-unittest 0.3.5, matching CI): 508/508 tests, 20/20 snapshots.


hungnguyen-anchore (approved):
Thanks for the update @holt!

PR #636, opened 2026-08-17
add a global image registry host override

blang9238:
## Summary

A customer mirroring the Anchore images into their own registry (Harbor) asked for a way to set the registry once instead of overriding every image stanza in `values.yaml`. This adds a single `global.imageRegistryHost` that re-homes every image the chart manages.

```yaml
global:
  imageRegistryHost: harbor.example.com
```

| Image value | Rendered as |
| --- | --- |
| `docker.io/anchore/enterprise:v6.1.1` | `harbor.example.com/anchore/enterprise:v6.1.1` |
| `bitnamilegacy/kubectl:1.30` (no host) | `harbor.example.com/bitnamilegacy/kubectl:1.30` |
| `alpine` (no host) | `harbor.example.com/alpine` |

## Semantics

Only the **registry host** is replaced; the repository, tag and digest on each image value are preserved. The leading path segment counts as a host when it contains a `.` or a `:`, or when it is `localhost` — the same rule the docker/OCI reference parsers use. Values with no host are prefixed rather than rewritten.

The override always wins over the per-image value. Every image value in these charts has a non-empty chart default, so "explicit value wins" would have made the key a permanent no-op — there is no way to tell a chart default from a user-supplied value. The per-image value still controls repository, tag and digest.

Both the string and dict image forms are handled. A dict that omits `registry` now renders successfully when the global host is set, instead of hitting the existing validation `fail`.

## Why `imageRegistryHost` and not `imageRegistry`

Helm merges `global` into every subchart, and the bundled redis and prometheus charts read `.Values.global.imageRegistry` as a **string**:

- `redis/charts/common/templates/_images.tpl:17`
- `prometheus/charts/kube-state-metrics/templates/_helpers.tpl:144`
- `prometheus/charts/prometheus-node-exporter/templates/daemonset.yaml:220`

Reusing that key as a map would have rendered their images as `map[host:harbor.example.com]/redis:7.4.6` — silently broken the moment `ui-redis` or `prometheus` is enabled.

## Scope

| Chart | Images covered | Version |
| --- | --- | --- |
| enterprise | `image`, `ui.image`, `kubectlImage`, `cloudsql.image`, `scratchVolume.fixerInitContainerImage` | 4.1.2 → 4.1.3 |
| k8s-inventory | `image.repository` | 0.6.4 → 0.6.5 |
| ecs-inventory | `image` | 0.0.17 → 0.0.18 |
| anchore-admission-controller | `image`, `initCa.image` | 0.8.4 → 0.8.5 |

All patch bumps — no application version changed.

`cloudsql.image` and `scratchVolume.fixerInitContainerImage` were rendering raw in `_common.tpl` and now go through the shared image helper.

**Out of scope, by design:** the bundled `ui-redis` and `prometheus` subcharts keep their own image values, and images supplied in `initContainers` / `extraInitContainers` are passed through to the pod spec verbatim. Both are called out in the README.

## Testing

Full render of the enterprise chart with `global.imageRegistryHost: harbor.bluestaq.com`, with cloudsql, the scratch-volume mode-fixer, the OSAA migration job and the upgrade job's kubectl init container all enabled:

```
harbor.bluestaq.com/anchore/enterprise@sha256:5d6111…
harbor.bluestaq.com/anchore/enterprise-ui@sha256:0d0ced…
harbor.bluestaq.com/bitnamilegacy/kubectl:1.30
harbor.bluestaq.com/cloudsql-docker/gce-proxy:1.37.8
harbor.bluestaq.com/alpine
docker.io/redis:7.4.6                                  <- subchart, untouched
```

helm-unittest: **614 enterprise / 5 k8s-inventory / 6 ecs-inventory / 5 anchore-admission-controller**, all passing. The 614 includes 16 new enterprise cases plus a small `image_validation_test.yaml`.

k8s-inventory, ecs-inventory and anchore-admission-controller had **no** helm-unittest suites at all. The unittest workflow iterates every changed chart directory, so the enterprise `Chart.yaml` bump would have failed that job on those three charts; each now has a suite covering the new value.

## Notes for reviewers

- CI installs these charts with `global.imageRegistryHost: ""` (a no-op), since pointing a `ci/` values file at a fake registry would just break image pulls. The unit tests are the real coverage.
- Because `anchore-admission-controller` and `k8s-inventory` are touched, `ENTERPRISE_REQUIRED=true` fires in `test.yaml` and each matrix leg stands up an extra enterprise install from the *released* chart. Expect a slow run.
- Unrelated pre-existing issue found while verifying: `helm template` on the enterprise chart fails on `gt .Values.anchoreConfig.analyzer.layer_cache_max_gigabytes 0.0` with "incompatible types for comparison" when the value is supplied via `--set`. It reproduces on a clean `main` and does not affect helm-unittest or normal installs.

### DIFF
--- a/stable/anchore-admission-controller/Chart.yaml
+++ b/stable/anchore-admission-controller/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: anchore-admission-controller
-version: 0.8.4
+version: 0.8.5
 appVersion: 0.8.4
 description: A kubernetes admission controller for validating and mutating webhooks that operates against Anchore Engine to make access decisions and annotations
 home: https://github.com/anchore/kubernetes-admission-controller
--- a/stable/anchore-admission-controller/templates/_helpers.tpl
+++ b/stable/anchore-admission-controller/templates/_helpers.tpl
@@ -23,6 +23,44 @@ Create chart name and version as used by the chart label.
 {{- end -}}
 
 {{/*
+Returns the global image registry host override, or an empty string when it isn't set.
+*/}}
+{{- define "anchore-admission-controller.globalImageRegistryHost" -}}
+{{- $global := default dict .Values.global -}}
+{{- default "" $global.imageRegistryHost | trimSuffix "/" -}}
+{{- end -}}
+
+{{/*
+Strips the registry host from a string image reference, returning everything after it.
+The leading path segment is treated as a registry host when it contains a '.' or a ':', or when it
+is 'localhost' - the same rule the docker/OCI reference parsers use. References that don't include a
+registry host (eg. 'anchore/kubernetes-admission-controller') are returned unchanged.
+*/}}
+{{- define "anchore-admission-controller.imageWithoutRegistryHost" -}}
+{{- $ref := trim . -}}
+{{- $parts := splitList "/" $ref -}}
+{{- $host := first $parts -}}
+{{- if and (gt (len $parts) 1) (or (contains "." $host) (contains ":" $host) (eq $host "localhost")) -}}
+{{- join "/" (rest $parts) -}}
+{{- else -}}
+{{- $ref -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Renders an image reference with its registry host replaced by global.imageRegistryHost when that is
+set. Accepts: dict "image" <image value> "context" <root context>
+*/}}
+{{- define "anchore-admission-controller.renderImage" -}}
+{{- $globalHost := include "anchore-admission-controller.globalImageRegistryHost" .context -}}
+{{- if $globalHost -}}
+{{- printf "%s/%s" $globalHost (include "anchore-admission-controller.imageWithoutRegistryHost" .image) -}}
+{{- else -}}
+{{- .image | trim -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Common labels
 */}}
 {{- define "anchore-admission-controller.labels" -}}
--- a/stable/anchore-admission-controller/templates/deployment.yaml
+++ b/stable/anchore-admission-controller/templates/deployment.yaml
@@ -44,7 +44,7 @@ spec:
       {{- end }}
       containers:
       - name: {{ .Chart.Name }}
-        image: "{{ .Values.image }}"
+        image: "{{ include "anchore-admission-controller.renderImage" (dict "image" .Values.image "context" .) }}"
         imagePullPolicy: {{ .Values.imagePullPolicy }}
       {{- with .Values.containerSecurityContext }}
         securityContext: {{- toYaml . | nindent 10 }}
--- a/stable/anchore-admission-controller/templates/init-ca/init-ca-hook.yaml
+++ b/stable/anchore-admission-controller/templates/init-ca/init-ca-hook.yaml
@@ -32,7 +32,7 @@ spec:
       {{- end }}
       containers:
       - name: create-ca
-        image:  "{{ .Values.initCa.image }}"
+        image:  "{{ include "anchore-admission-controller.renderImage" (dict "image" .Values.initCa.image "context" .) }}"
       {{- with .Values.containerSecurityContext }}
         securityContext: {{- toYaml . | nindent 10 }}
       {{- end }}
--- a/stable/anchore-admission-controller/templates/init-ca/init-ca-script.yaml
+++ b/stable/anchore-admission-controller/templates/init-ca/init-ca-script.yaml
@@ -15,9 +15,18 @@ data:
     export PATH=$PATH:/tmp/.bin
     cd /tmp/.bin
 
-    curl -L -o jq https://github.com/jqlang/jq/releases/download/jq-1.7.1/jq-linux-amd64
+    # The tools below are fetched per-architecture. Downloading the amd64 builds on an
+    # arm64 node passes the `which` checks further down but fails with an exec format
+    # error the first time kubectl actually runs.
+    case "$(uname -m)" in
+      x86_64)        ARCH=amd64 ;;
+      aarch64|arm64) ARCH=arm64 ;;
+      *) echo "unsupported architecture: $(uname -m)"; exit 1 ;;
+    esac
+
+    curl -L -o jq https://github.com/jqlang/jq/releases/download/jq-1.7.1/jq-linux-${ARCH}
     chmod +x jq
-    curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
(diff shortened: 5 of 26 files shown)


ekkjerstad:
Supporting a path in the value (`harbor.example.com/anchore`) is exactly our layout, and having it
covered in the unit tests is appreciated.

We modelled `global.imageRegistryHost` against our live `enterprise` values before adopting it, and
hit two problems in the new code path plus one merge blocker. Reporting them rather than filing
new feature requests — all three are within the scope the PR already claims.

## 1. An explicitly-set image value is silently discarded when the global is set

`enterprise.renderImage` rewrites the host on every image value, with no way to exempt one:

- string form — the leading host is stripped and the global is prepended, whether or not the
  operator set that value deliberately
- dict form — `merge (dict "registry" $globalHost) $image` gives the global precedence, so it
  overwrites a per-image `registry`

So an override of one of the images the PR lists as covered:

```yaml
global:
  imageRegistryHost: harbor.example.com/anchore-licensed
kubectlImage: harbor.example.com/dockerhub/bitnamilegacy/kubectl:1.30
```

renders as:

```
harbor.example.com/anchore-licensed/dockerhub/bitnamilegacy/kubectl:1.30
```

Our prefix is treated as just another host to strip. There is no render-time error — the result is
a plausible-looking reference that 404s at pull time, which is difficult to diagnose without
reading the helper.

The reason we can't simply pick one prefix is Anchore's own licensing model. `anchore/enterprise`
and `anchore/enterprise-ui` are entitlement-gated, so they have to resolve through a proxy-cache
project holding our licensee credentials. The public helper images the chart also deploys
(`bitnamilegacy/kubectl`, `alpine`) must not use those credentials, so they resolve through a
separate, generic Docker Hub proxy-cache project. Two upstream credential sets means two mirror
prefixes, and the feature currently supports one with no escape hatch — which makes it unusable
for us in either direction:

| `imageRegistryHost` set to | Result |
| --- | --- |
| the licensee project | `image` / `ui.image` correct; `kubectlImage` / `fixerInitContainerImage` resolve to a project that cannot serve them |
| the generic project | the reverse — the two entitlement-gated images resolve to a project with no entitlement |

We're not asking for a particular design here. Any mechanism that lets an explicitly-set value
survive the global would resolve it. One cheap option, since a template can't distinguish
"operator set this" from "this is the chart default" without comparing against the shipped default:

```yaml
kubectlImage:
  registry: harbor.example.com/dockerhub
  repository: bitnamilegacy/kubectl
  tag: "1.30"
  ignoreGlobalRegistryHost: true
```

## 2. Host-less defaults render to unpullable paths under any mirror

Two defaults have no registry host, and one of them also omits its Docker Hub namespace:

```yaml
kubectlImage: bitnamilegacy/kubectl:1.30
scratchVolume:
  fixerInitContainerImage: alpine
```

`imageWithoutRegistryHost` correctly leaves these unchanged and prefixes them, so `alpine` becomes
`<prefix>/alpine`. That path doesn't exist in a Docker Hub mirror. The implicit-`library`
namespace is a Docker Hub *client* convenience — the actual repository path is `library/alpine`,
and a Harbor proxy-cache project requires it in full. So `fixerInitContainerImage`, which the PR
lists as covered, cannot be made to resolve for any mirrored install.

Rather than special-casing implicit-`library` in the helper, the smaller fix looks like making the
defaults fully qualified:

```yaml
kubectlImage: docker.io/bitnamilegacy/kubectl:1.30
scratchVolume:
  fixerInitContainerImage: docker.io/library/alpine:3.21
```

That makes strip-host-and-reprefix behave uniformly for every image value, removes the host-less
branch as a correctness dependency, and is a no-op for installs pulling directly from Docker Hub.

## 3. The chart version bump collides with `main`

The PR bumps `stable/enterprise` 4.1.2 → 4.1.3, but `main` is already at 4.1.3, so this will need a
rebase and a bump to 4.1.4.

## Not raising

The bundled `ui-redis` and `prometheus` subcharts are documented as out of scope and we're fine
setting those ourselves — noting it only so it's clear we're not asking for it.

For what it's worth, `k8s-inventory` works for us exactly as implemented, since it pulls from a
single upstream. Both problems above are specific to `enterprise`, where the image set spans two
upstreams with different credentials.

hungnguyen-anchore:
Hey @blang9238   thanks for the PR. Regarding the implementation of this, I'd like to propose an alternative solution (I actually thought through a few). The main reasoning behind this was im of the camp that explicitly set registries in the values should always take precedence, otherwise, with the current implementation, theres no way to set a different registry host.

The first thing i thought through was what if we made all images hostless, then have the default be docker.io, then they could override it. If set at the values level though, we wouldn't replace a host from a key. This had some drawbacks though. eg:

You have to restate the whole reference to exempt one image. If you want `kubectlImage` from a different mirror, the only way to say so is to write out the full `harbor.example.com/dockerhub/bitnamilegacy/kubectl:1.30` — which pins the tag too. So you've (the user) now taken ownership of the version and we (the chart) can never bump it for you. Eg. if we bump kubectl to 1.32, the user who exempted kubectl image will have to see/know to bump their pin to 1.32.

Also gcr.io/cloudsql-docker/gce-proxy was an issue because if it was hostless, it'd default to docker.io which doesn't exist.

Eventually, i ended on: What if we changed all the images to dict type. (eg change the string image defaults to their dict variant with registry, repository, and tag or digest.

With this, an image value that has no registry defined will use the `global.imageRegistryHost`. If it does have a registry, then we'd keeps it. This would let us do the top level imageRegistryHost for main things, then override it at each image level for things like thirdparty images, eg (using the anchore-licensed and dockerhub proxy cache path example):

  ```yaml
  global:
    imageRegistryHost: harbor.example.com/anchore-licensed   # image, ui.image
  kubectlImage:
    registry: harbor.example.com/dockerhub
  scratchVolume:
    fixerInitContainerImage:
      registry: harbor.example.com/dockerhub
  cloudsql:
    image:
      registry: harbor.example.com/gcr
  ```

which would render:

  ```
  harbor.example.com/anchore-licensed/anchore/enterprise@sha256:5d61…
  harbor.example.com/anchore-licensed/anchore/enterprise-ui@sha256:0d0c…
  harbor.example.com/dockerhub/bitnamilegacy/kubectl:1.30
  harbor.example.com/dockerhub/library/alpine:3.21
  harbor.example.com/gcr/cloudsql-docker/gce-proxy:1.37.8
  ```

I think that would also cover @ekkjerstad case without needing the additional `ignoreGlobalRegistryHost` flag. For `cloudsql.image` we could put a default of `registry: gcr.io` since it isn't a Docker Hub image, so it's exempt by default and you override it like above.

A complete image reference string would still works and is used as written, so existing values files are unaffected. A string with no host takes the global like any other value. We can keep the host detection the same for string type image values (eg dot / colon / `localhost`).

I've pushed this to `hn/registry-override-structured` (https://github.com/anchore/anchore-charts/pull/638) . Happy to discuss this more, and get ya'lls thoughts on it.

blang9238:
Closing as OBE based on https://github.com/anchore/anchore-charts/pull/638.

PR #638, opened 2026-08-24
Global Registry Host support, image type updated from string to dict

hungnguyen-anchore:


blang9238:
**Legacy string image values in nested paths log a `cannot overwrite table with non table` warning**

The dict conversion stays backward compatible — an existing values file that pins a string reference still renders correctly, I checked all five. But the nested image values now emit a Helm warning on every render, while the top-level ones don't:

```
$ helm template t . --set-string ui.image=docker.io/anchore/enterprise-ui:v9.9.9
coalesce.go:298: warning: cannot overwrite table with non table for enterprise.ui.image
  (map[digest:sha256:0d0ceda… registry: repository:anchore/enterprise-ui tag:])
```

| value set as a string | warning | renders correctly |
| --- | --- | --- |
| `image` | no | yes |
| `kubectlImage` | no | yes |
| `ui.image` | yes | yes |
| `scratchVolume.fixerInitContainerImage` | yes | yes |
| `cloudsql.image` | yes | yes |

Purely cosmetic — the value is used in every case. The concern is that anyone upgrading with a values file that pins `ui.image` sees this on every `helm install` and `helm upgrade`, and "cannot overwrite" reads like their setting was discarded when it wasn't. A line in the release notes would probably head off the support tickets: the string is still honored, and setting `registry` / `tag` instead avoids the warning.

I don't have an explanation for why the top-level values are exempt — the warning fires per path inside Helm's coalesce and only the nested ones trip it.

blang9238:
**`cloudsql.image` being exempt by default is the right call, but it's silent**

Shipping `registry: gcr.io` is correct — leaving it empty would render `docker.io/cloudsql-docker/gce-proxy` on a default install, which doesn't exist — and the split-mirror example in the README does set it.

The sharp edge is that the exemption is invisible at render time. An operator mirroring everything sets `global.imageRegistryHost`, sees a fully re-homed install, and doesn't discover the proxy is still pointed at `gcr.io` until they enable cloudsql — at which point a cluster that can't reach `gcr.io` just fails to pull, with nothing in the chart output suggesting why. Every other image follows the global, so this is the one exception nobody's looking for.

Not asking for a behavior change. A short "mirroring everything" checklist that enumerates the complete set of values a fully-mirrored install has to touch — the global, `cloudsql.image.registry`, plus anything passed through `initContainers` / `extraInitContainers` — would likely be enough. All the pieces are already in the doc; what's missing is the "and that's the whole list" framing.

blang9238 (dismissed):
- Validation test gap (nit)

The one failure test covers "nothing supplies a registry". A dict with repository but neither tag nor digest hits the same fail and isn't covered. One extra case.

- Per-image registry doesn't get its trailing slash trimmed — the global does (low-moderate)

enterprise.globalImageRegistryHost applies trimSuffix "/", but default $globalHost $image.registry doesn't, so a per-image registry with a trailing slash renders an invalid reference:

```
--set kubectlImage.registry=harbor.example.com/dockerhub/
→ harbor.example.com/dockerhub//bitnamilegacy/kubectl:1.30
```

The global tolerates the same input fine. // isn't a legal path component in an OCI reference, so this fails at pull time with an unhelpful error. It matters more than a normal typo-tolerance nit because the README actively steers split-mirror users toward setting per-image registry — that's the documented path for exactly the customer's case. One-line fix: trimSuffix "/" on $registry after the default.

blang9238 (dismissed):
LGTM!

anchore-jesset (dismissed):
LGTM

anchore-jesset (approved):

PR #641, opened 2026-08-28
bug fix: Osaa migration service account

hungnguyen-anchore:
Setting a serviceAccountName on the OSAA migration job gave the pod an account the chart never granted scale-down RBAC to. The upgrade job bound its Role to a different, generated name and the migration job skipped RBAC creation entirely, so the scale-down was denied, and because the script discarded kubectl's exit code, it reported success while the migration ran against services that were never scaled down.

Now we bind the Role to whichever ServiceAccount the pod actually runs as, and the scale-down lists the deployments first so it fails the hook with kubectl's exit code when denied, while still passing cleanly when there are genuinely no deployments to scale (eg. if the user did a kubectl delete deployments before they use helm to bring back the deployments).


gnyahay (approved):
Pretty sure I ran into this a few times.

anchore-jesset (approved):
lgtm!

